PR #65, opened 2026-03-31
fix(core): deterministic node ID normalization for parallel batch output

### LINKED ISSUE #59: Parallel file-analyzer subagents can produce inconsistent node IDs and invalid complexity values
## Problem

When `/understand --full` dispatches parallel file-analyzer subagents, there is no deterministic enforcement of node ID format or complexity enum values. The prompt specifies the correct formats, but the assembly pipeline trusts LLM output without validation — so inconsistent batches silently corrupt the final graph.

## Issue 1: No runtime enforcement of node ID format

The file-analyzer prompt (`skills/understand/file-analyzer-prompt.md`, lines 219–227) specifies:

| Node Type | Required Format | Example |
|---|---|---|
| File | `file:<relative-path>` | `file:src/index.ts` |
| Function | `func:<relative-path>:<name>` | `func:src/utils.ts:formatDate` |
| Class | `class:<relative-path>:<name>` | `class:src/models/User.ts:User` |

However, the Zod schema only validates `id: z.string()` (`packages/core/src/schema.ts`, line 13) — any string passes. Neither Phase 3 (ASSEMBLE) nor the `GraphBuilder` (`packages/core/src/analyzer/graph-builder.ts`, line 84) validates ID prefix format on merged batch output.

Since subagents are LLMs writing JSON directly to `batch-<N>.json` files, they can produce:
- Project-name-prefixed IDs: `myproject:backend/main.py`
- Double-prefixed IDs: `myproject:service:docker-compose.yml`
- Bare paths with no prefix: `frontend/src/utils/constants.ts`

**Evidence from a 226-file project run:**
```jsonc
// Batch 1 — correct
{ "id": "file:backend/app/api/audit.py" }

// Batch 4 — project name prefixed
{ "id": "noora-health-res-cms:backend/main.py" }

// Batch 8 — no prefix
{ "id": "frontend/src/utils/constants.ts" }
```

## Issue 2: Complexity enum not validated during assembly

The schema defines `complexity: z.enum(["simple", "moderate", "complex"])` (`packages/core/src/schema.ts`, line 20). The `parseFileAnalysisResponse` function (`packages/core/src/analyzer/llm-analyzer.ts`, lines 97–116) normalizes invalid values to `"moderate"` — but this function is only used in the programmatic `GraphBuilder` path.

The `/understand` skill's subagents write batch JSON files directly to disk. Phase 3 merges these files without passing nodes through `parseFileAnalysisResponse`, so invalid complexity values survive into the assembled graph:

```jsonc
// LLM used "low"/"high" instead of "simple"/"complex"
{ "id": "file:backend/app/api/__init__.py", "complexity": "low" }

// LLM used numeric scale
{ "id": "file:backend/app/services/eval.py", "complexity": 4 }

// LLM used free-text
{ "id": "document:CLAUDE.md", "complexity": "detailed" }
```

## Issue 3: Cascading edge drops at assembly

Phase 3 (`skills/understand/SKILL.md`, line 155) removes edges whose `source` or `target` references a non-existent node ID. When Issue 1 causes ID mismatches between what edges reference and what nodes were actually saved as, valid edges get silently dropped:

```jsonc
// Edge references correct format, but the node was saved with wrong ID format
{ "source": "file:backend/app/services/turnio_client.py", "target": "file:backend/app/core/config.py", "type": "imports" }
// Actual node ID was "noora-health-res-cms:file:backend/app/services/turnio_client.py" → mismatch → dropped
```

## Issue 4: All-or-nothing dashboard validation

The dashboard calls `validateGraph()` on load (`packages/dashboard/src/App.tsx`, lines 122–130). Since this is Zod `safeParse`, **any** invalid node (e.g., `complexity: "low"`) causes the **entire graph** to fail to load with an error banner. There is no partial load or per-node auto-correction.

This means if even one batch produces a non-enum complexity value and it survives to the final JSON, the dashboard won't render at all.

## Impact

On a 226-file project, we observed: nodes with 3 different ID formats across batches, non-standard complexity values (`"low"`, `"high"`, numeric, free-text), and layers with depleted node counts due to ID mismatches during assembly.

## Root Cause

The assembly pipeline (Phase 3 in `SKILL.md`) trusts that all subagent batches produce consistent output. The only format enforcement is:
1. **Prompt instructions** (not deterministic)
2. **Graph reviewer** in Phase 6 (also LLM-based, not deterministic)
3. **Zod schema** at final write/load (catches invalid values but rejects the whole graph rather than fixing them)

There is no deterministic normalization step between batch merge and final write.

## Suggested Fixes

1. **Add a normalization pass in Phase 3 (ASSEMBLE):** After merging batches, deterministically normalize all node IDs to `type:path` format (strip project-name prefixes, add missing `file:` prefixes) and map complexity values to the valid enum (`low`→`simple`, `high`→`complex`, numeric/other→`moderate`).

2. **Validate node ID prefix format in the Zod schema:** Change `id: z.string()` to `id: z.string().regex(/^(file|func|class|module|concept):/)` so invalid IDs are caught at validation time rather than silently passing.

3. **Run `parseFileAnalysisResponse`-style normalization on batch JSON:** Before merging batch files in Phase 3, pass each node through the same complexity normalization that `llm-analyzer.ts` already implements.

fishinakleinbottle:
Resolves #59

## Problem

Parallel file-analyzer subagents can produce inconsistent node IDs (project-name prefixed, double-prefixed, bare paths) and numeric complexity values. The existing `autoFixGraph`/`COMPLEXITY_ALIASES` pipeline handles string alias correction but does not address malformed IDs or numeric complexity — causing cascading edge drops at assembly and potential dashboard load failures.

## What this PR adds

A pre-assembly normalization pass that runs **before** the existing sanitize/autoFix/validate pipeline, handling the things it doesn't cover.

### Changed files

| File | What changed |
|------|-------------|
| `packages/core/src/analyzer/normalize-graph.ts` | **New.** `normalizeNodeId` (strips project-name/double prefixes, adds missing type prefixes), `normalizeComplexity` (maps numeric 1-10 scale), `normalizeBatchOutput` (orchestrates ID fix → edge rewriting → dedup) |
| `packages/core/src/__tests__/normalize-graph.test.ts` | **New.** 31 tests covering all normalization paths |
| `packages/core/src/index.ts` | Exports the new module |
| `packages/core/src/__tests__/schema.test.ts` | 1 test confirming schema accepts bare string IDs (documents the design decision that normalization, not schema, enforces prefix format) |
| `skills/understand/SKILL.md` | Rewrites Phase 3 ASSEMBLE from 3 bullets to a 6-step normalization sequence |
| `skills/understand/file-analyzer-prompt.md` | Adds a warning box about ID format requirements |

### Not changed

- **Zod schema** — `id` stays as `z.string()`. Normalization fixes IDs upstream of validation.
- **Dashboard** — no changes needed.
- **String complexity aliases** — already handled by `COMPLEXITY_ALIASES` in `autoFixGraph`.

## Reviewer checklist

- [ ] `normalize-graph.ts` — `stripToValidPrefix()` loop: verify it correctly peels project-name prefixes, double-prefixes, and bottoms out on bare paths
- [ ] `normalize-graph.ts` — `normalizeBatchOutput()`: confirm edge rewriting happens *after* all node IDs are normalized (order matters)
- [ ] `SKILL.md` Phase 3: check the 6-step sequence aligns with the code logic
- [ ] Verify no overlap with `autoFixGraph`/`sanitizeGraph`/`normalizeGraph` in `schema.ts` — this PR intentionally avoids duplicating that work

## Test plan

- [x] `pnpm --filter @understand-anything/core build` — passes
- [x] `pnpm --filter @understand-anything/core test` — 189 tests pass
- [ ] Run `/understand --full` on a multi-file project and verify no dropped nodes/edges in dashboard

### DIFF
--- a/understand-anything-plugin/packages/core/src/__tests__/normalize-graph.test.ts
+++ b/understand-anything-plugin/packages/core/src/__tests__/normalize-graph.test.ts
@@ -4,6 +4,7 @@ import {
   normalizeComplexity,
   normalizeBatchOutput,
 } from "../analyzer/normalize-graph.js";
+import { validateGraph } from "../schema.js";
 
 describe("normalizeNodeId", () => {
   it("passes through a correct file ID unchanged", () => {
@@ -83,12 +84,12 @@ describe("normalizeNodeId", () => {
     ).toBe("concept:caching");
   });
 
-  it("handles double project-name prefix", () => {
+  it("handles project-name prefix before a valid non-code prefix", () => {
     expect(
       normalizeNodeId("my-project:service:docker-compose.yml", {
         type: "file",
       }),
-    ).toBe("file:docker-compose.yml");
+    ).toBe("service:docker-compose.yml");
   });
 
   it("returns empty string for empty input", () => {
@@ -97,6 +98,26 @@ describe("normalizeNodeId", () => {
 
   it("falls back to untouched ID for unknown node type", () => {
     expect(normalizeNodeId("some-id", { type: "widget" as any })).toBe("some-id");
+  });
+
+  it("passes through non-code type IDs unchanged", () => {
+    expect(normalizeNodeId("config:tsconfig.json", { type: "config" })).toBe("config:tsconfig.json");
+    expect(normalizeNodeId("document:README.md", { type: "document" })).toBe("document:README.md");
+    expect(normalizeNodeId("service:docker-compose.yml", { type: "service" })).toBe("service:docker-compose.yml");
+    expect(normalizeNodeId("table:migrations/001.sql:users", { type: "table" })).toBe("table:migrations/001.sql:users");
+    expect(normalizeNodeId("endpoint:src/routes.ts:GET /api/users", { type: "endpoint" })).toBe("endpoint:src/routes.ts:GET /api/users");
+    expect(normalizeNodeId("pipeline:.github/workflows/ci.yml", { type: "pipeline" })).toBe("pipeline:.github/workflows/ci.yml");
+    expect(normalizeNodeId("schema:schema.graphql", { type: "schema" })).toBe("schema:schema.graphql");
+    expect(normalizeNodeId("resource:main.tf", { type: "resource" })).toBe("resource:main.tf");
+  });
+
+  it("adds prefix for bare paths with non-code types", () => {
+    expect(normalizeNodeId("tsconfig.json", { type: "config" })).toBe("config:tsconfig.json");
+    expect(normalizeNodeId("README.md", { type: "document" })).toBe("document:README.md");
+  });
+
+  it("strips project-name prefix from non-code type IDs", () => {
+    expect(normalizeNodeId("my-project:config:tsconfig.json", { type: "config" })).toBe("config:tsconfig.json");
   });
 });
 
@@ -257,6 +278,13 @@ describe("normalizeBatchOutput", () => {
 
     expect(result.edges).toHaveLength(0);
     expect(result.stats.danglingEdgesDropped).toBe(1);
+    expect(result.stats.droppedEdges).toHaveLength(1);
+    expect(result.stats.droppedEdges[0]).toEqual({
+      source: "file:src/a.ts",
+      target: "file:src/nonexistent.ts",
+      type: "imports",
+      reason: "missing-target",
+    });
   });
 
   it("deduplicates nodes keeping last occurrence", () => {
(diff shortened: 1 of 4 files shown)


Lum1104 (commented):
In `normalize-graph.ts`, I think the edge rewrite is still a bit too strict for the inconsistency this PR is trying to handle.

  Right now source/target are rewritten only through exact idMap lookup from the original node IDs. That works when the edge uses the same malformed ID string as the node did, but it misses cases where parallel outputs produce different malformed variants for the same logical node.

  Example:

  - node ID: src/bare.ts -> normalizes to file:src/bare.ts
  - edge source: my-project:file:src/bare.ts

  Those refer to the same node, but since the edge string does not exactly match the node’s original id, it won’t be rewritten and will be dropped as dangling.

  I think the bigger issue is that this becomes a silent relationship loss: once the edge is dropped here, later validation no longer sees a dangling reference to report, so the graph just loses the connection with only the aggregate counter incremented.

  Would it make sense to canonicalize edge endpoints through the same normalization logic before the dangling-edge check, or otherwise emit enough detail here for the caller to surface which edges were dropped?

PR #85, opened 2026-04-13
Refactor/improve graph builder

avenue9977:
refactor: performance and correctness improvements to GraphBuilder 

What changed and why                                                                                                      
- **O(1) duplicate detection** - replaced the per-call `new Set(this.nodes.map(n => n.id))` rebuild (O(n²) overall) with a `nodeIds` class field updated incrementally at every insertion
- **Edge deduplication** - `addImportEdge` and `addCallEdge` now skip duplicate edges via an `edgeKeys` set,  preventing inflated edge counts when multiple agents report the same relationship
-  **Safe `build()` output** - `nodes` and `edges `are returned as shallow copies, preventing callers from mutating builder internals                                                                          
-  **`addChildNode` helper** - consolidated the repeated dedup-check + node-push + contains-edge pattern  across all five child-node loops in `addNonCodeFileWithAnalysis`
-  **KIND_TO_NODE_TYPE constant** - moved kind→node type mapping to module level, allocated once instead of on every call 
-  **basename helper** - extracted repeated `filePath.split("/").pop()` into a `private static` method      
-  **`addNonCodeFile` returns its ID** - eliminates the independent fileId re-computation that could silently drift                                                                                      
-  **`localeCompare` in language sort** - ensures reliable alphabetical ordering      

P.S. Great project! Happy to contribute!

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -13,13 +13,13 @@ importers:
         version: 5.9.3
       vitest:
         specifier: ^3.1.0
-        version: 3.2.4(@types/debug@4.1.12)(@types/node@25.5.0)(jiti@2.6.1)(lightningcss@1.31.1)(yaml@2.8.3)
+        version: 3.2.4(@types/debug@4.1.12)(@types/node@25.5.0)(jiti@2.6.1)(lightningcss@1.32.0)(yaml@2.8.3)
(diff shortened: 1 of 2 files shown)


Lum1104:
Thanks! @codex review this

PR #96, opened 2026-04-21
fix: expand ${PLUGIN_DIR} with double quotes in hook echo commands

xiaolai:
> **Automated audit**: This PR was generated by [NLPM](https://github.com/xiaolai/nlpm-for-claude), a natural language programming linter, running via [claude-code-action](https://github.com/anthropics/claude-code-action). Please evaluate the diff on its merits.

## Bug

In `understand-anything-plugin/hooks/hooks.json`, both the `PostToolUse` and `SessionStart` hook commands use single-quoted strings for the `echo` message that instructs Claude to read `${PLUGIN_DIR}/hooks/auto-update-prompt.md`. In bash, **single quotes suppress all variable expansion**, so `${PLUGIN_DIR}` is passed literally as the five characters `$`, `{`, `P`, `L`, `U`, ... instead of being resolved to the actual plugin directory path.

If Claude Code's hook runner does not pre-expand `${PLUGIN_DIR}` before handing the command to bash, Claude will receive a path like `${PLUGIN_DIR}/hooks/auto-update-prompt.md` — which does not exist — and silently skip the auto-update instructions.

## Fix

Changed the `echo` argument from single quotes to double quotes in both hooks:

```diff
- echo '[understand-anything] ... ${PLUGIN_DIR}/hooks/auto-update-prompt.md ...'
+ echo "[understand-anything] ... ${PLUGIN_DIR}/hooks/auto-update-prompt.md ..."
```

This is the minimal one-character change per hook. Double quotes allow `${PLUGIN_DIR}` to expand at runtime, while still preserving the surrounding bash command structure. No other logic is changed.

## Impact

Without this fix, the auto-update feature (both commit-triggered and session-start staleness check) silently sends Claude a non-existent file path, meaning the knowledge graph is never incrementally updated even when `autoUpdate: true` is configured.

### DIFF
--- a/understand-anything-plugin/hooks/hooks.json
+++ b/understand-anything-plugin/hooks/hooks.json
@@ -6,7 +6,7 @@
         "hooks": [
           {
             "type": "command",
-            "command": "echo \"$TOOL_INPUT\" | grep -qE 'git\\s+(commit|merge|cherry-pick|rebase)' && [ -f .understand-anything/config.json ] && grep -q '\"autoUpdate\".*true' .understand-anything/config.json && [ -f .understand-anything/knowledge-graph.json ] && echo '[understand-anything] Commit detected with auto-update enabled. You MUST read the file at ${PLUGIN_DIR}/hooks/auto-update-prompt.md and execute its instructions to incrementally update the knowledge graph. Do not ask the user for confirmation — just do it.' || true"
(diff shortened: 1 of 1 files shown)


Lum1104:
Thanks for catching this @xiaolai — single quotes blocking expansion is a real bug.

Closing in favor of 1e07cba, which makes one additional change: the variable name itself was wrong. Per the [Claude Code hooks docs](https://code.claude.com/docs/en/hooks), the injected variable is ${CLAUDE_PLUGIN_ROOT}, not ${PLUGIN_DIR} (which Claude Code doesn't inject — it would have expanded to an empty string after this PR's quote fix). The rest of the codebase already uses ${CLAUDE_PLUGIN_ROOT} consistently in skills/understand/SKILL.md etc., so this aligns the hook with that.

Credited you as co-author on the commit. Appreciate the audit!

PR #108, opened 2026-04-27
feat(dashboard): add read-only source viewer

### LINKED ISSUE #107: Add read-only local source viewer to dashboard
## Problem

The dashboard is useful for understanding how files, functions, classes, and layers relate, but when a user wants to inspect the actual source behind a selected node they still have to leave the dashboard and jump back to an editor.

This is especially noticeable after selecting a function/class node: the graph knows the file path and often has a line range, but the dashboard currently does not provide a read-only way to inspect that source in context.

Related to #94 in spirit: that issue describes the friction of having to jump back into source files. This proposal is narrower than Mermaid/logic diagrams: show the actual local source file, read-only, from the dashboard.

## Proposed solution

Add a small read-only source viewer to the dashboard:

- Add a `Files` tab in the right sidebar, built from `graph.nodes[].filePath`.
- Let users double-click an analyzed file to open its source.
- Add an `Open code` action to node details when a node has a `filePath`.
- Reuse the existing `CodeViewer` surface for actual source text rather than node metadata.
- Show syntax highlighting, line numbers, and highlighted node line ranges when available.
- Add a wider modal/expanded viewer so code is readable without overhauling the graph layout.

## Local source endpoint

Because the dashboard already runs as a local Vite server against a project `GRAPH_DIR`, source can be fetched on demand without storing source text in the graph:

`GET /file-content.json?token=<token>&path=<relative-path>`

Proposed guardrails:

- require the same dashboard token
- resolve paths under the same `GRAPH_DIR` project root
- reject absolute paths, `..` traversal, null bytes, directories, binary files, and oversized files
- only serve files that are already present in the knowledge graph
- do not store file contents in `.understand-anything/knowledge-graph.json`
- show a graceful "source unavailable" state in static/demo mode

## Why this feels useful

- Keeps the "understand the codebase" workflow inside the dashboard longer.
- Helps onboarding and review prep because users can move from graph structure to source without context switching.
- Keeps scope small: read-only viewer, no editor, no Monaco/VS Code clone.
- Avoids graph schema changes.

## Implementation status

I have a working local branch and can open a PR if this direction is welcome.

Tested locally with:

- `pnpm --filter @understand-anything/skill test`
- `pnpm --filter @understand-anything/dashboard build`
- `pnpm --filter @understand-anything/dashboard build:demo`

### 'Files' tab:

Image

### Code view (double click on filename in pane):

Image

### Expanded view (read only):

Image

arkaigrowth:
## Summary

Closes #107.

Adds a small read-only source viewer to the dashboard so users can move from the graph to actual local source code without leaving the dashboard.

Key pieces:

- add a `Files` tab in the right sidebar, built from analyzed `graph.nodes[].filePath` values
- allow double-clicking an analyzed file to open its source
- add `Open code` in `NodeInfo` when the selected node has a `filePath`
- repurpose `CodeViewer` to render actual source text with line numbers, Prism highlighting, horizontal scroll, and node line-range highlighting
- add an expanded modal viewer so code is readable without redesigning the graph layout
- add a token-protected local file endpoint:
  - `GET /file-content.json?token=<token>&path=<relative-path>`

## Security / local-file behavior

The endpoint is intentionally narrow:

- requires the dashboard token
- resolves paths only under the active `GRAPH_DIR`
- rejects absolute paths, `..` traversal, null bytes, directories, binary-looking files, and files over the size limit
- only serves files already present in the knowledge graph
- does not write file contents into `.understand-anything/knowledge-graph.json`
- returns a graceful "source unavailable" state in static/demo contexts

## Why this shape

This intentionally avoids becoming an editor or VS Code clone. The goal is a quick read-only inspection loop:

graph node -> source file -> relevant lines -> back to graph

This also differs from the old bottom code viewer that was removed in #56: that viewer duplicated node metadata, while this one fetches and renders actual source text on demand.

### 'Files' tab:

Image

### Code view (double click on filename in pane):

Image

### Expanded view (read only):

Image

## Test plan

- [x] `git diff --check`
- [x] `pnpm --filter @understand-anything/core build`
- [x] `pnpm --filter @understand-anything/skill build`
- [x] `pnpm --filter @understand-anything/core test` (654 tests)
- [x] `pnpm --filter @understand-anything/skill test` (696 tests)
- [x] `pnpm --filter @understand-anything/dashboard build`
- [x] `pnpm --filter @understand-anything/dashboard build:demo`
- [x] Manual smoke: open dashboard, Files tab, double-click `apps/telegram-control-plane/src/App.tsx`, open expanded modal
- [x] Endpoint smoke: valid file succeeds; missing token returns `403`; traversal returns `400`

### DIFF
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,7 +1,7 @@
 {
   "name": "understand-anything",
   "description": "AI-powered codebase understanding — analyze, visualize, and explain any project",
-  "version": "2.3.2",
+  "version": "2.4.0",
   "author": {
     "name": "Lum1104"
   },
--- a/.copilot-plugin/plugin.json
+++ b/.copilot-plugin/plugin.json
@@ -1,7 +1,7 @@
 {
   "name": "understand-anything",
   "description": "AI-powered codebase understanding — analyze, visualize, and explain any project",
-  "version": "2.3.2",
+  "version": "2.4.0",
   "author": {
     "name": "Lum1104"
   },
--- a/.cursor-plugin/plugin.json
+++ b/.cursor-plugin/plugin.json
@@ -2,7 +2,7 @@
   "name": "understand-anything",
   "displayName": "Understand Anything",
   "description": "AI-powered codebase understanding — analyze, visualize, and explain any project",
-  "version": "2.3.2",
+  "version": "2.4.0",
   "author": {
     "name": "Lum1104"
   },
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -20,8 +20,8 @@ An open-source tool combining LLM intelligence + static analysis to produce inte
 - Dark luxury theme: deep blacks (#0a0a0a), gold/amber accents (#d4a574), DM Serif Display typography
 - Graph-first layout: 75% graph + 360px right sidebar
 - No ChatPanel or Monaco Editor
-- Sidebar: ProjectOverview (default) → NodeInfo (node selected) → LearnPanel (Learn persona)
-- Code viewer: styled summary overlay (slides up from bottom on file node click)
+- Sidebar tabs: `Info` (ProjectOverview default → NodeInfo when node selected → LearnPanel in Learn persona, composing) and `Files` (FileExplorer tree built from the structural graph)
+- Code viewer: prism-react-renderer source viewer that slides up from the bottom on file node click; an expand button promotes it into a full-screen modal. Source content is fetched from the dev server's `/file-content.json` endpoint, gated by access token + a graph-derived path allowlist
 - Schema validation on graph load with error banner
 
 ## Agent Pipeline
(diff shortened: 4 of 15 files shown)


arkaigrowth:
I added screenshots showing the sidebar file tree/code viewer and the expanded modal.

Marking as ready for review, but I’d appreciate feedback on the local file endpoint/security boundary before this is mergeable. tysm!

Lum1104:
Sounds good. I was wondering if you support the diff mode~

arkaigrowth:
> Sounds good. I was wondering if you support the diff mode~

Good Q. Right now this PR works alongside the existing diff overlay, but it does not yet render Git-style diff hunks.

if `diff-overlay.json` is loaded, the existing dashboard diff mode still marks changed/affected nodes, and opening one of those nodes in the source viewer shows the current local source with the node’s existing `lineRange` highlighted.

I can add lightweight diff-mode support to this PR by surfacing changed/affected state in the Files tab and CodeViewer header... and optionally adding a changed-only filter...

Tbh I’d probably keep full hunk-level diff rendering as a follow-up b/c the current diff overlay looks like it contains node IDs rather than old/new source text or changed line ranges... so that'd need a larger endpoint/API design.

Lum1104:
Cool, let's have that for the future features. @codex review this

PR #2, opened 2026-03-17
Diff Mode

KyleWMiller:
## Summary

  Adds a visual diff overlay to the knowledge graph dashboard that highlights changed and affected components when reviewing PRs or branch diffs.

### Preview
image

  ### What it does

  - **Changed nodes**: Red ring + glow for files directly modified
  - **Affected nodes**: Amber ring + glow for 1-hop connected components (blast radius)
  - **Faded nodes**: Unrelated components fade out when diff mode is active
  - **Edge highlighting**: Connections between diff-related nodes are emphasized, others fade
  - **Toggle**: "Diff ON/OFF" button in the header (mirrors existing Layers toggle pattern)
  - **Auto-detect**: Dashboard auto-activates diff mode when `diff-overlay.json` is present

  ### How it works

  1. Run `/understand-anything:understand-diff` on any feature branch. It automatically diffs against `main` (or specify a base branch in the prompt, e.g.
  `/understand-anything:understand-diff against develop`).
  2. The skill writes `.understand-anything/diff-overlay.json` alongside the knowledge graph after analyzing the branch diff.
  3. Launch the dashboard with `/understand-anything:understand-dashboard`. The Vite middleware serves `diff-overlay.json` at `/diff-overlay.json`.
  4. On load, `App.tsx` fetches the overlay data and populates Zustand store state.
  5. `GraphView.tsx` and `CustomNode.tsx` apply conditional styling based on diff state.
  6. Composes cleanly with existing features (layers, search, tour, personas).

  ### Files changed

  | File | Change |
  |------|--------|
  | `src/index.css` | Diff color CSS variables + glow/fade utility classes |
  | `src/store.ts` | `diffMode`, `changedNodeIds`, `affectedNodeIds` state + actions |
  | `src/components/DiffToggle.tsx` | **New** - toggle button + color legend component |
  | `src/components/CustomNode.tsx` | Diff-aware node styling (changed/affected/faded) |
  | `src/components/GraphView.tsx` | Diff state wired into node data + conditional edge colors |
  | `src/App.tsx` | Fetches `diff-overlay.json` on load, renders DiffToggle |
  | `vite.config.ts` | Middleware to serve `diff-overlay.json` |
  | `skills/understand-diff/SKILL.md` | New step 8: write diff overlay after analysis |

  ### `diff-overlay.json` schema

  ```json
  {
    "version": "1.0.0",
    "baseBranch": "main",
    "generatedAt": "2026-03-16T19:00:00Z",
[...]

PR #124, opened 2026-05-06
fix(skill): make /understand work on Windows + pnpm 10

tipich:
## Summary

Fixes two bugs that block `/understand` from running properly on common modern dev setups (Windows + pnpm 10):

1. **`extract-structure.mjs:34,37` — Windows ESM URL bug.** `await import()` was called with raw absolute paths returned by `require.resolve()` and `path.resolve()`. On Windows those start with `C:\...`, which Node 24's ESM loader parses as URL scheme `C:` and rejects with `ERR_UNSUPPORTED_ESM_URL_SCHEME`. Wrap both with `pathToFileURL().href` so the loader receives a proper `file://` URL. POSIX systems are unaffected (their absolute paths start with `/`, which Node ESM accepts).

2. **`pnpm.onlyBuiltDependencies` missing from `package.json`.** pnpm 10 (released ~2024-09) changed the default to block all postinstall scripts unless explicitly approved. Without this allowlist, `pnpm install` skips the native build for tree-sitter parsers, esbuild, and sharp — which means `registry.analyzeFile()` returns `undefined` and any code path that calls it (including `buildFingerprintStore`) crashes. Cross-platform issue, not Windows-specific.

## Why both matter

Without these, on a clean Windows + pnpm 10 install:

- file-analyzer subagents can't run `extract-structure.mjs` → fall back to manual file reads → `importMap` stays empty across all batches → assemble-reviewer has to grep-recover hundreds of cross-batch edges by hand
- `buildFingerprintStore` crashes → no `fingerprints.json` baseline gets generated → the **incremental update path described in `SKILL.md` Phase 0 step 7 doesn't work at all**, every `/understand` is a full rebuild

## Verification

End-to-end on a 1190-file Laravel codebase:

- After fix #1: `extract-structure.mjs` resolves `@understand-anything/core` cleanly via `pathToFileURL`
- After fix #2: tree-sitter PHP parser ships during `pnpm install` (visible `node-gyp-build` Done lines)
- `buildFingerprintStore(ROOT, 1190 files, registry, hash)` builds the baseline in 1.56s
- `extractFileFingerprint` correctly extracts 18 typed functions from `Booking.php`
- Synthetic change detection correctly classifies whitespace-only diff as COSMETIC and added-method diff as STRUCTURAL

## Test plan

- [ ] Smoke `/understand` on a small Windows project to confirm extract-structure no longer falls back
- [ ] Confirm `pnpm install` on a fresh Windows + pnpm 10 clone produces compiled tree-sitter parsers without manual `pnpm approve-builds`
- [ ] Confirm `buildFingerprintStore` no longer crashes when called per the pattern in `SKILL.md` Phase 7 step 2.5

### DIFF
--- a/understand-anything-plugin/package.json
+++ b/understand-anything-plugin/package.json
@@ -15,5 +15,22 @@
     "@types/node": "^22.0.0",
     "typescript": "^5.7.0",
     "vitest": "^3.1.0"
+  },
+  "pnpm": {
+    "onlyBuiltDependencies": [
+      "esbuild",
+      "sharp",
+      "tree-sitter-c",
+      "tree-sitter-c-sharp",
+      "tree-sitter-cpp",
+      "tree-sitter-go",
+      "tree-sitter-java",
+      "tree-sitter-javascript",
+      "tree-sitter-php",
+      "tree-sitter-python",
+      "tree-sitter-ruby",
+      "tree-sitter-rust",
+      "tree-sitter-typescript"
+    ]
   }
 }
--- a/understand-anything-plugin/pnpm-lock.yaml
+++ b/understand-anything-plugin/pnpm-lock.yaml
@@ -100,6 +100,18 @@ importers:
       devlop:
         specifier: ^1.1.0
         version: 1.1.0
+      elkjs:
+        specifier: ^0.9.3
+        version: 0.9.3
+      graphology:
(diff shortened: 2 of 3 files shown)


Lum1104:
@codex review this

Lum1104 (approved):

PR #156, opened 2026-05-17
fix(pipeline): handle ALTER migrations and auto-init fingerprint registry

tipich:
## Summary

Two small, unrelated gaps in `/understand` that both surface on Laravel projects. Bundled into one PR because the diff is tiny and the two fixes are independent — easy to back out either side.

### Gap 1 — ALTER migrations produce orphan `table:` nodes

`agents/file-analyzer.md` only described how to handle CREATE-shape SQL migrations. A Laravel `Schema::table('users', ...)` migration (and the equivalent Rails `change_table` / Django `AddField`) emitted a bare `file:` node with **no `migrates` edge to the existing `table:` node** it was modifying. Each ALTER added an orphan sibling table node to the graph.

Measured impact on one Aeropax (Laravel 13) regen: 8 orphan ALTER-migration `table:` nodes after a full incremental pipeline pass.

**Fix:** added an explicit *ORM migration files* section to the Non-code edge creation guidance, distinguishing CREATE shape (`Schema::create`, `create_table`, `CreateModel`) from ALTER shape (`Schema::table`, `change_table`, `AddField`/`RemoveField`/`AlterField`). The ALTER case emits **only** the `migrates` edge to the existing table node — no duplicate `table:` node. Also added two rows to the Edge Signal Quick Reference table for fast pattern matching.

### Gap 2 — `buildFingerprintStore` required hidden setup the skill doc didn't mention

`skills/understand/SKILL.md` Phase 7 step 2.5 documents:

```js
import { buildFingerprintStore, saveFingerprints } from '@understand-anything/core';

const store = await buildFingerprintStore('<PROJECT_ROOT>', sourceFilePaths);
saveFingerprints('<PROJECT_ROOT>', store);
```

…but the actual signature was `(projectDir, filePaths, registry, gitCommitHash)` — sync, with both extra args required. Following the doc gave:

```
TypeError: Cannot read properties of undefined (reading 'analyzeFile')
    at packages/core/dist/fingerprint.js:168:35
```

…because `registry` was `undefined`. Result: `fingerprints.json` silently failed to regen on Phase 7, leaving the baseline stale.

**Fix:** option (A) from the proposal — auto-init inside `buildFingerprintStore`.

- `registry` and `gitCommitHash` are now optional.
- When `registry` is missing OR empty, the function constructs one and registers **all built-in non-code parsers** (`registerAllParsers`) **plus** a `TreeSitterPlugin` with `builtinLanguageConfigs` (the 10 supported code languages, WASM grammars awaited).
- When `gitCommitHash` is missing, the function shells out to `git rev-parse HEAD` in `projectDir`; falls back to `\"unknown\"` if not a git repo or git is unavailable.
- The function is now `async` to match the documented `await` and to allow the tree-sitter WASM init. Callers that already pass a populated registry are untouched — auto-registration only fires on an empty registry.

Updated SKILL.md to call out that the 2-argument form is the supported standalone path and that 3rd/4th args remain available for explicit-control / CI scenarios.

## Test plan

- [x] `pnpm --filter @understand-anything/core build` clean (no TS errors).
- [x] `pnpm --filter @understand-anything/core test` — **673 tests, all green** (previously 670; added 3 new in `fingerprint-autoinit.test.ts`).
- [x] New test file uses the **real filesystem** (the existing `fingerprint.test.ts` mocks `node:fs`, so the smoke test had to live separately). It covers:
  - Zero-arg call returns a populated store (`{ version: \"1.0.0\", files: { ... }, gitCommitHash: \"unknown\" }` outside a git repo).
  - Tree-sitter extracts `function`/`class` signatures for both TypeScript and Python sample files (this is the direct regression test for the `analyzeFile undefined` crash).
  - Caller-provided non-empty registry stays untouched — auto-registration is skipped, sentinel plugin output wins, explicit commit hash flows through.
- [ ] Aeropax regen (real-project smoke) — after this lands, re-running `/understand` against `C:\dev\aeropax` should drop orphan ALTER-migration count from 8 → 0 and produce a fresh `fingerprints.json`.

## Pre-existing failure

`pnpm --filter @understand-anything/skill test` has one pre-existing failure in `extract-structure.test.mjs` (1 fail / 43 pass on `main`; 1 fail / 44 pass on this branch). Verified on `main` — unrelated to these changes.

## Out of scope

- Other orphan classes from the same Aeropax regen (36 `config()` runtime-string orphans, 24 markdown→code document orphans, 268 legacy file orphans) — bigger investigations, separate tasks.
- Adding a Laravel framework prompt to `skills/understand/frameworks/` — the new guidance in `file-analyzer.md` is framework-agnostic and covers Rails / Django / Laravel uniformly. A dedicated `laravel.md` can come later if it justifies its own prompt.

### DIFF
--- a/understand-anything-plugin/agents/file-analyzer.md
+++ b/understand-anything-plugin/agents/file-analyzer.md
@@ -298,6 +298,18 @@ The `batchImportData` values contain only resolved project-internal paths — ex
 - **Doc files:** If the doc mentions specific files, components, or modules by name, create `documents` edges. README.md typically documents the project entry point.
 - **Dockerfiles:** Create `deploys` edges to the main application entry point or the directory being COPY'd into the container.
 - **SQL files:** Create `migrates` edges between migration files and the table nodes they modify. Create `defines_schema` edges from schema files to API handlers that serve that data.
+- **ORM migration files (Laravel, Rails, Django, etc.):** PHP/Ruby/Python migration files that manage schema are `code` fileCategory but behave like SQL migrations. They modify tables but do not always create them. Distinguish two cases:
+  - **CREATE shape** (creates a new table): emit a `table:<path>:<name>` node AND a `migrates` edge from the file node to that table node (weight 0.7).
+    - Laravel: `Schema::create('users', ...)`
+    - Rails: `create_table :users`
+    - Django: `migrations.CreateModel(name='User', ...)`
+  - **ALTER shape** (modifies an existing table): emit ONLY a `migrates` edge from the file node to the existing `table:<existing-path>:<name>` node (weight 0.7). Do NOT create a new `table:` node — that would create a duplicate sibling of the original CREATE migration's table node.
+    - Laravel: `Schema::table('users', ...)`, `Schema::rename('old', 'new')`, `Schema::drop('users')`
+    - Rails: `change_table :users`, `add_column :users, ...`, `remove_column :users, ...`, `rename_table :old, :new`
+    - Django: `migrations.AddField(model_name='user', ...)`, `migrations.RemoveField(...)`, `migrations.AlterField(...)`, `migrations.RenameModel(...)`
+  - **Resolving the target `table:` node id:** prefer the original CREATE migration's path if you can identify it from the batch input or from referenced project files. If you only have the table name and cannot resolve the original path, use the convention `table:<this-migration-path>:<name>` and rely on a later pass to merge — but PREFER linking to the existing node when possible. The merge step normalizes IDs but does not invent links, so a `migrates` edge to a slightly mis-IDed table node is still better than a bare file node with no edge.
+  - **Multi-table migrations:** a single migration file may touch several tables. Emit one `migrates` edge per touched table.
+  - **Down migrations:** treat the `down`/`reverse` method the same — it migrates the same tables in reverse. Do not emit separate edges for it.
 - **CI configs:** Create `triggers` edges to the deployment targets or test suites they invoke.
 - **GraphQL/Protobuf schemas:** Create `defines_schema` edges to the code files that implement the resolvers or service handlers.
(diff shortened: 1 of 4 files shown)


Lum1104:
Thanks for the PR! Status differs between the two gaps, could you split it?

Gap 1 (ALTER migrations / orphan table: nodes) — please keep. agents/file-analyzer.md on main still has no ORM-migration guidance, so your CREATE/ALTER distinction + the two Quick Reference rows are still needed. The Aeropax measurement (8 orphan ALTER-migration nodes) is great signal. Rebase should be clean.

Gap 2 (buildFingerprintStore Phase 7 2.5) — please drop. I fixed the underlying bug in 2.7.3 via commit e7af9ae (fix(skills/understand): bundle build-fingerprints.mjs and reorder Phase 7), but with a different approach: I kept buildFingerprintStore's public API as-is (sync, 4 required args) and bundled a skills/understand/build-fingerprints.mjs that does await TreeSitterPlugin.init() + PluginRegistry wiring + the 4-arg call correctly; Phase 7 now invokes that script and writes fingerprints before meta.json. So your fingerprint.ts API change, the new fingerprint-autoinit.test.ts, and the SKILL.md 3-line diff would now duplicate (and conflict with) that bundled-script path — please drop them.

TL;DR: shrink the PR to just the agents/file-analyzer.md diff, rebase on main, and I'll merge. Thanks!

tipich:
Done — thanks for the clear split.

- **Gap 1 (ALTER migrations):** kept. Only the `agents/file-analyzer.md` diff remains (the CREATE/ALTER distinction + the two Quick Reference rows).
- **Gap 2 (buildFingerprintStore):** dropped — confirmed `e7af9ae` is in `main` and `skills/understand/build-fingerprints.mjs` is bundled, so the `fingerprint.ts` API change, `fingerprint-autoinit.test.ts`, and the SKILL.md diff are gone to avoid duplicating/conflicting with the bundled-script path.

Rebased on `main`; applied cleanly and the PR now shows MERGEABLE (1 file, +14/-0). Ready when you are.

PR #161, opened 2026-05-21
chore: add ESLint tooling with TypeScript support

Kartalops:
## Summary

The repository has a pnpm lint script that runs eslint . but ESLint and typescript-eslint were not listed in devDependencies, causing the script to fail.

Added:
- eslint (^9.0.0)
- @eslint/js (^9.0.0)
- typescript-eslint (^8.0.0)

Also added eslint.config.mjs with flat config format (ESLint 9+) using typescript-eslint strict rules. Ignores node_modules, dist, build, and framework-specific output directories.

## Problem

Running pnpm lint fails with eslint: not found because ESLint is not installed as a dependency.

## Solution

Add the required ESLint packages as devDependencies and create the flat-config file. Does not fix any linting errors — this PR only adds the tooling infrastructure.

## Risk / Compatibility

- No runtime code changes
- No new runtime dependencies added
- Only tooling (devDependency) additions

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,6 +20,9 @@ jobs:
       - name: Install dependencies
         run: pnpm install
 
+      - name: Lint
+        run: pnpm lint
+
       - name: Build core
         run: pnpm --filter @understand-anything/core build
 
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -1,0 +1,29 @@
+import eslint from '@eslint/js';
+import tseslint from 'typescript-eslint';
+
+export default tseslint.config(
+  eslint.configs.recommended,
+  ...tseslint.configs.strict,
+  {
+    files: ['**/*.ts', '**/*.tsx', '**/*.mts', '**/*.cts'],
+    languageOptions: {
+      parserOptions: {
+        projectService: {
+          allowDefaultProject: ['*.config.*'],
+        },
+      },
+    },
+  },
+  {
+    ignores: [
+      '**/node_modules/**',
+      '**/dist/**',
+      '**/build/**',
+      '**/.understand-anything/**',
+      '**/.claude-plugin/**',
+      '**/.cursor-plugin/**',
+      '**/.copilot-plugin/**',
+      '**/coverage/**',
+    ],
+  },
+);
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -1,29 +1,56 @@
 import eslint from '@eslint/js';
 import tseslint from 'typescript-eslint';
+import globals from 'globals';
 
 export default tseslint.config(
-  eslint.configs.recommended,
-  ...tseslint.configs.strict,
-  {
-    files: ['**/*.ts', '**/*.tsx', '**/*.mts', '**/*.cts'],
-    languageOptions: {
-      parserOptions: {
-        projectService: {
-          allowDefaultProject: ['*.config.*'],
-        },
-      },
-    },
-  },
   {
     ignores: [
       '**/node_modules/**',
       '**/dist/**',
       '**/build/**',
+      '**/public/**',
+      '**/coverage/**',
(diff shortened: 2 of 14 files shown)


Lum1104:
Thanks for picking this up! A few notes before we land it:

On the Copilot comments

1. Lockfile — agreed, this needs fixing. CI runs pnpm install on GitHub Actions where CI=true is set automatically, and pnpm defaults to --frozen-lockfile in that case. Please run pnpm install locally and commit the updated pnpm-lock.yaml.
2. ignores position — I think Copilot is wrong on this one. In ESLint 9 flat config, a config object containing only an ignores key (no files, no rules, etc.) is treated as global ignores regardless of its position in the array. See the ESLint docs and the typescript-eslint examples — they put ignores last too. Your block has only ignores, so it works as intended. No change needed here.

A few things I'd like to sort out before merging

3. Trailing newline — eslint.config.mjs is missing a newline at EOF (the diff shows \ No newline at end of file). Minor, but eol-last would catch it.
4. projectService in a pnpm workspace — have you actually run pnpm lint end-to-end? With type-aware rules enabled and allowDefaultProject limited to *.config.*, I'm worried files under packages/core and packages/dashboard will hit parserOptions.project errors because each package has its own tsconfig.json and root-level eslint . has to figure out which one each file belongs to. If you've already confirmed it works, great — otherwise we may need to either drop type-aware linting or wire up per-package tsconfig resolution.
5. strict vs recommended — since this is the first time the repo is being linted, tseslint.configs.strict is almost certainly going to produce hundreds of errors. The PR description says "does not fix any linting errors," which I read as "lint will exit non-zero from day one." Two options I'd prefer:
  - Start with tseslint.configs.recommended so we get a useful baseline that's closer to green, then ratchet up to strict in a follow-up.
  - Or keep strict but include enough fixes (or rule overrides) in this PR that pnpm lint actually exits 0.
6. CI wiring — .github/workflows/ci.yml doesn't currently run pnpm lint. If we want this to enforce anything, we should add the step. Happy to do that in a follow-up if you'd rather keep this PR scoped to tooling — just want to flag it so we don't end up with installed-but-unused tooling.

Let me know which direction you want to take on (5) and I'll review the next push.

Lum1104:
@okwn pushed a follow-up commit addressing the review notes; hope you don't mind me taking it the rest of the way. Summary of what changed:

**Direction picked for (5):** went with `tseslint.configs.recommended` baseline + fixed everything so `pnpm lint` exits 0. Ratcheting up to `strict` is a clean follow-up PR.

**Config (`eslint.config.mjs`)**
- `strict` → `recommended`.
- Dropped the `parserOptions.projectService` block. Neither `recommended` nor `strict` is type-aware, so it was unused; removing it also dodges the pnpm-workspace tsconfig-resolution failure mode I was worried about in (4).
- Added Node + browser globals via the `globals` package — fixes 280 `no-undef` errors in `.mjs` scripts and dashboard code.
- Expanded ignores: `**/public/**` (kills 1900+ errors in the homepage demo bundle), `**/.astro/**`, `.private/**`.
- Allow `_`-prefixed unused vars; `no-irregular-whitespace` with `skipComments: true` (json-parser intentionally uses ZWSP-escaped block-comment examples in JSDoc).
- EOF newline added (3).

**Code fixes (the 13 residual real errors)**
- Dropped unused imports/vars in `ruby-extractor.ts`, `tree-sitter-plugin.ts`, two parsers, two test files, and `change-classifier.ts`.
- Replaced two `as any[]` in `schema.ts` with `Array<Record<string, unknown>>` — the runtime checks already narrowed properly.
- Dropped unused catch binding in `extract-structure.mjs`.

**Other**
- Refreshed `pnpm-lock.yaml` (1).
- Wired `pnpm lint` into `.github/workflows/ci.yml` (6).

Local verification: `pnpm lint` exits 0, core + skill builds pass, all 1445 tests pass. Waiting on CI now.

PR #176, opened 2026-05-22
feat: --hybrid flag routes file-analyzer batches to local Gemma via Ollama (~4-5× cost reduction)

Itsthewayofyou:
## Summary

- Adds `--hybrid` flag to `/understand` skill that routes the expensive **file-analyzer** extraction phase to a local Gemma model via Ollama instead of Claude subagents
- Introduces `hybrid_runner.py` — a self-contained Python orchestrator that handles tree-sitter extraction + Gemma semantic analysis + schema-safe post-processing
- Reduces Claude API cost by **~4-5×** on medium/large repos. Architecture-reasoning phases (architecture-analyzer, tour-builder, domain-analyzer) stay on Claude.

## How it works

```bash
# Standard (unchanged)
/understand

# Hybrid — file-analyzer batches go to local Gemma
/understand --hybrid

# Configure the model/endpoint
OLLAMA_MODEL=gemma4:26b-a4b OLLAMA_HOST=http://localhost:11434 /understand --hybrid
```

**Routing table:**

| Phase | Standard | Hybrid |
|-------|----------|--------|
| project-scanner | Claude | Claude (low cost, dynamic script) |
| file-analyzer ×N batches | Claude | **Gemma via Ollama** |
| assemble-reviewer | Claude | Claude (semantic judgment) |
| architecture-analyzer | Claude | Claude |
| tour-builder | Claude | Claude |
| domain-analyzer | Claude | Claude |

## Key design decisions

**Why only file-analyzer?** It's ~80% of total token cost — runs in 5 concurrent batches of 20-30 files each. The other agents are either low-cost or require Claude-level reasoning quality.

**Import edge injection**: Gemma tends to re-resolve imports from source instead of strictly using `batchImportData`. `hybrid_runner.py` strips all Gemma-emitted `imports` edges and reinserts them deterministically from `batchImportData` (pre-resolved by project-scanner). This makes import graphs identical to standard mode regardless of model quality.

**Graceful fallback**: If Gemma fails (timeout, parse error), the runner writes an empty `batch-N.json`. The pipeline continues; assemble-reviewer notes the gap and merge-batch-graphs.py handles it.

**Safe for existing users**: `HYBRID_MODE=false` by default. No behavior change unless `--hybrid` is explicitly passed.

## Requirements

- Ollama running locally with a capable model (tested: `gemma4:26b-a4b`, any model ≥13B should work)
- Python 3 (stdlib only — no new dependencies)
- The `extract-structure.mjs` script already bundled with the skill runs unchanged

## Tested on

- `gemma4:26b-a4b` via Ollama
- Calibration: 100% node ID prefix correctness, 100% edge type vocabulary, JSON validity ✅
- Output flows correctly through `merge-batch-graphs.py` normalization

## Test plan

- [ ] `python3 hybrid_runner.py --help` shows scan/analyze subcommands
- [ ] Run `/understand --hybrid` on a small repo (20-50 files) — verify `batch-*.json` written
- [ ] Compare assembled graph node/edge counts between standard and hybrid modes
- [ ] Verify `OLLAMA_MODEL` env var respected
- [ ] Verify Ollama-down fallback logs warning and continues


Lum1104:
Sounds good, can this run faster or it just reduce the token usage in your use case?

@codex review this

simkimsia:
if  i increase the concurrent batches from 5 to 10, will taht speed things up?

> Why only file-analyzer? It's ~80% of total token cost — runs in 5 concurrent batches of 20-30 files each. The other agents are either low-cost or require Claude-level reasoning quality.

Itsthewayofyou:
@Lum1104 Good question — it's primarily a cost play, not a speed one. The ~4-5× is token/API cost, not wall-clock. On most setups it's actually a touch *slower*: the 5 file-analyzer batches all hit a single local Ollama instance, so a large model on one GPU serializes generation, whereas standard mode runs those batches with Claude's cloud parallelism. So: meaningfully cheaper, comparable-to-slightly-slower on speed depending on your local hardware. If latency (not cost) is the goal, standard mode is still the faster path.

Also worth noting the fix I just pushed (`a84b942`): a connection-refused/DNS failure now correctly falls back to an empty batch instead of crashing the run, and a node emitted without a `type` no longer gets silently dropped at validation.

Itsthewayofyou:
@simkimsia In **hybrid** mode, no — and it can make things slightly worse. All batches queue against a single Ollama endpoint, so throughput is bound by the GPU and `OLLAMA_NUM_PARALLEL`, not by how many batches you fan out. Sending 10 instead of 5 just deepens the queue and adds memory/context pressure on a large model. The levers that actually move hybrid speed are the local model size/quant, `OLLAMA_NUM_PARALLEL`, and your GPU.

In **standard** (Claude) mode, raising concurrency *does* help — up to your API rate limits — since each batch is an independent cloud request.

PR #185, opened 2026-05-23
fix(windows): make install.ps1 PowerShell 5.1 friendly

Bortlesboat:
## Summary

- make `install.ps1` parse reliably in Windows PowerShell 5.1 by keeping the no-BOM script ASCII-only
- preserve every current platform, including newer Vibe, nanobot, and Kiro support
- document native Windows update and uninstall commands alongside the Unix commands

## Verification

- `powershell.exe -NoProfile -ExecutionPolicy Bypass -File ./install.ps1 -Help`
- Windows PowerShell parser API: zero parse errors
- non-ASCII sweep of `install.ps1`: zero matches
- `git diff origin/main...HEAD --check`

## Refresh note

Rebuilt on current `main` as `3369980`. The obsolete Vibe addition from the original draft was dropped because upstream already added it; newer Kiro configuration and Codex invocation guidance are retained.


Bortlesboat:
Refreshed onto current main in 3369980 and marked ready. The conflict resolution preserves newer platforms and Kiro/Codex guidance, drops the duplicate Vibe addition, removes every remaining non-ASCII character from install.ps1, and keeps the Windows lifecycle docs. Native Windows PowerShell help and parser checks pass with zero non-ASCII matches and a clean diff check.

PR #197, opened 2026-05-24
fix(dashboard): restrict Path Finder "To" dropdown to reachable nodes

### LINKED ISSUE #188: FEATURE Path filter
https://understand-anything.com/demo/#:~:text=Export-,Path,-Theme

PATH 2 drop down lists 1st available nods ! when one selected ! Only keep in 2nd list those paths that actually work. Easy to scan edges and preload only working paths in second drop down list.

amazing work BTW :D

ShresthSamyak:
## Summary

- Compute the set of nodes reachable from the chosen **From** node and filter the **To** dropdown to that set, so the user can't pick a destination with no path. Closes #188.
- Memoize the undirected adjacency on the graph so reachability and `findPath` agree and the work isn't redone on every render.
- Polish: disable **To** until a source is selected, show the reachable count next to the **To** label, and clear a stale **To** selection if the user changes **From** to a source from which the previous target is unreachable.

## Before

Both selects listed every node. Pairs with no connecting path only failed *after* clicking **Find Path**, requiring the user to manually scan edges to find working destinations.

## After

Once **From** is picked, **To** only shows nodes actually reachable from it — `N reachable` is shown next to the label. Selecting **To** before **From** is disabled with a placeholder hinting at the order. If the source has no neighbors, **To** shows `No reachable nodes from source` instead of empty options.

## Notes

- Reachability uses the same bidirectional traversal as the existing `findPath` (treats graph edges as undirected) so the filter and the BFS result can never disagree.
- The "No path found" branch in the result panel is now defensively unreachable for valid inputs but is left in place since it's cheap.

## Test plan

- [x] `pnpm install`
- [x] `pnpm --filter @understand-anything/dashboard exec tsc --noEmit` — clean
- [x] `pnpm --filter @understand-anything/dashboard build` — succeeds
- [x] `pnpm --filter @understand-anything/dashboard test` — 42 passed
- [x] `pnpm lint` — clean
- [ ] Manually verify in the dashboard demo: pick a **From** node, confirm **To** only lists reachable destinations and the count matches.
- [ ] Verify **To** is disabled until **From** is picked, and that changing **From** to a node that can't reach the prior target clears the **To** selection.

### DIFF
--- a/understand-anything-plugin/packages/dashboard/src/components/PathFinderModal.tsx
+++ b/understand-anything-plugin/packages/dashboard/src/components/PathFinderModal.tsx
@@ -1,4 +1,4 @@
-import { useEffect, useRef, useState } from "react";
+import { useEffect, useMemo, useRef, useState } from "react";
 import { useDashboardStore } from "../store";
 
 interface PathFinderModalProps {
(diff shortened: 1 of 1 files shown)


Lum1104:
@codex review this

PR #201, opened 2026-05-24
Fix pnpm detection for macOS zsh installs

sashishjha:
## Summary
Fixes a macOS/zsh setup issue where pnpm can be installed but unavailable to the Bash tool used by `/understand` and `/understand-dashboard`.

## Changes
- Add PNPM_HOME/PATH bootstrap before pnpm commands
- Try Corepack when pnpm is unavailable
- Improve missing-pnpm error message
- Use `pnpm exec vite` for dashboard launch
- Document the macOS/zsh edge case in README

## Testing
- Ran `git diff --check`
- Ran `bash -n install.sh`
- Full test suite not run; change is limited to skill instructions and README documentation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -202,6 +202,7 @@ The installer clones the repo to `~/.understand-anything/repo` and creates the r
 - Supported `<platform>` values: `gemini`, `codex`, `opencode`, `pi`, `openclaw`, `antigravity`, `vibe`, `vscode`, `hermes`, `cline`, `kimi`
 - Update later: `./install.sh --update`
 - Uninstall: `./install.sh --uninstall <platform>`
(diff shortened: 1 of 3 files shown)


Lum1104:
Thanks for the PR, @sashishjha, and for the thorough write-up.

I'm going to pass on this one. The root cause is a local shell-configuration issue, not a plugin bug: pnpm setup on macOS writes PNPM_HOME to ~/.zprofile, which is only sourced by login zsh shells. The fix lives on the user's side — source ~/.zprofile once, or restart the terminal/IDE after installing pnpm — and from that point on pnpm is on PATH like any other CLI. This is exactly how pnpm's own installer documents it.

I'd rather not work around it inside the skill prompts:

- It hard-codes macOS-specific paths ($HOME/Library/pnpm) into a cross-platform skill.
- It adds ~15 lines of shell preamble to two skill files for an environment quirk that's already solved upstream.
- Every invocation now pays the token/prompt-noise cost for an edge case most users never hit.
- The Corepack fallback would silently mask "pnpm not installed" vs. "installed but not on PATH" — the current error is more honest.

Closing this out. Appreciate the effort regardless. 🙏

Issue #214, opened 2026-05-25
extract-import-map: tsconfig paths targets with leading ./ (e.g. "@/*": ["./*"]) fail to resolve → cross-module import  edges dropped


TsungTang:
### Version
  Plugin 2.7.5 (Claude Code).

  ### Summary
  When a project's `tsconfig.json` declares path aliases whose **target starts with `./`** — e.g. the very common `"@/*":  ["./*"]` — `extract-import-map.mjs` fails to resolve those aliased imports. Relative (`./foo`) imports still resolve, so the  graph ends up with **only intra-directory edges**; all cross-module/cross-layer dependencies (the `@/...` imports) are  silently missing. This breaks the dependency view, `/understand-diff` blast-radius, and "what depends on X".

### Repro (minimal)

  ```jsonc
  // tsconfig.json
  { "compilerOptions": { "paths": { "@/*": ["./*"] } } }
  ```
  ```ts
  // app/page.tsx
  import { x } from "@/lib/thing"; // NOT resolved
  import { y } from "./local";     // resolved
  ```

  Run the scan/import-map → the `@/lib/thing` edge is absent.

  ### Matrix (only the `./` prefix matters, not baseUrl)
  | tsconfig `paths` | baseUrl | resolves? |
  |---|---|---|
  | `"@/*": ["./*"]` | (none) | ❌ |
  | `"@/*": ["./*"]` | `"."` | ❌ |
  | `"@/*": ["./*"]` | `"src"` | ❌ |
  | `"@/*": ["*"]` | (none) | ✅ |
  | `"@/*": ["*"]` | `"."` | ✅ |

  ### Root cause
  In `resolveTsJsImport` (skills/understand/extract-import-map.mjs), for a **root-level** tsconfig `tsConfigDir === ""`  (falsy), so `posix.join` is never called and the leading `./` in the mapped target is never normalized. The candidate becomes `"./lib/thing"` while `fileSet` holds the normalized `"lib/thing"`, so `probeWithExtensions` never matches. When `posix.join` *is* invoked (nested config dir or non-`.` baseUrl) it normalizes the `./` away — which is why those paths happen to differ, but the `./`-target + root-config case slips through.

  ### Proposed fix (one line)

  Normalize the candidate, e.g. strip the leading `./` from the mapped target:
  ```js
  const mapped = applyTsAlias(target, aliasMatch).replace(/^\.\//, '');
  ```
[...]

PR #239, opened 2026-05-26
fix(extract-import-map): resolve tsconfig paths with './' prefix (#214)

Ylsssq926:
Hit this on a Next.js project — all `@/` alias imports were missing from the knowledge graph. Traced it to `extract-import-map.mjs` line 433.

**Root cause:** When tsconfig uses `"@/*": ["./*"]` (the default in Next.js / Vite / Nuxt), `applyTsAlias` produces paths like `"./utils"`. The ternary expression at line 433 skipped `posix.join` when `tsConfigDir` was empty (root tsconfig), so the `"./"` prefix passed through to `probeWithExtensions` unchanged. Since `fileSet` stores paths without `"./"` (e.g. `"utils.ts"`), the probe never matched.

**Fix:** Remove the ternary and always run `posix.join(tsConfigDir, relativeToConfig)`. When `tsConfigDir` is `''`, `posix.join('', './utils')` normalizes to `'utils'` — matching fileSet. When `tsConfigDir` is non-empty, behavior is identical to before.

3 lines → 1 line change in the script, plus 3 new test cases:
- `"./*"` target (Next.js App Router default)
- `"./src/*"` target
- Multi-target fallback with `"./"` prefixes

Existing tests still pass — the `"src/*"` form (without `./`) was never affected.


Ylsssq926:
Rebased this branch onto current main to resolve the conflict. The net PR diff is now empty because current main already includes the #214 import-map fix and focused regression coverage for tsconfig path targets with leading ./ prefixes. Leaving the PR open for maintainer/author decision rather than closing it unilaterally.

Issue #162, opened 2026-05-21
bug: extract-structure.mjs isCli guard silently no-ops when invoked via symlinked SKILL_DIR


DonQuilatte:
## Summary

`skills/understand/extract-structure.mjs` silently exits 0 without writing output when invoked via a symlinked `SKILL_DIR` path. The bundled `build-fingerprints.mjs` is likely affected by the same pattern.

This is silently-failing — no error message, no exit code. Every subagent that follows the documented invocation thinks structural extraction succeeded but produces no output file.

## Reproduction

Steps:

1. Install the plugin via the documented Claude Code path. `~/.agents/skills/understand/` ends up as a symlink into the plugin checkout (e.g., `~/.understand-anything-plugin/skills/understand/` → `~/.understand-anything/repo/understand-anything-plugin/skills/understand/`).
2. From a project root, follow `SKILL.md` Phase 1 Step 2 verbatim:
   ```bash
   node ~/.agents/skills/understand/extract-structure.mjs \
     "$PROJECT_ROOT/.understand-anything/tmp/ua-file-analyzer-input-1.json" \
     "$PROJECT_ROOT/.understand-anything/tmp/ua-file-extract-results-1.json"
   ```
3. Observe: exit 0, no stderr, **no output file written.**

## Root cause

The `isCli` guard at the bottom of `extract-structure.mjs` compares `import.meta.url` against `pathToFileURL(process.argv[1]).href`:

- `import.meta.url` resolves the script's location through symlinks → returns the **realpath** URL.
- `pathToFileURL(process.argv[1]).href` preserves the symlink → returns the **symlinked-path** URL.

When the two paths differ, the equality check is `false` and `main()` never runs. The script exits normally with no work performed.

Worked around in our evaluation by invoking through the realpath:

```bash
node ~/.understand-anything/repo/understand-anything-plugin/skills/understand/extract-structure.mjs ...
```

In a recent evaluation of Understand-Anything on a Bash-heavy project, **6 of 6 parallel file-analyzer subagents independently hit this** before workarounds were applied. It was the single most common failure mode in the run.

## Suggested fix

Canonicalize both sides of the comparison with `realpathSync`:

```js
import { realpathSync } from 'node:fs';
import { fileURLToPath, pathToFileURL } from 'node:url';

function isCli() {
  try {
    const moduleRealPath = realpathSync(fileURLToPath(import.meta.url));
    const argvRealPath = realpathSync(process.argv[1]);
[...]

Issue #152, opened 2026-05-15
Bug: auto-update PARTIAL_UPDATE overwrites fingerprints.json instead of merging — causes permanent FULL_UPDATE regression


rilkerfc:
## Problem

After a successful `PARTIAL_UPDATE` run, `fingerprints.json` is overwritten with **only the re-analyzed batch** instead of being merged with the existing entries. All files not in that batch lose their fingerprints permanently.

On the next auto-update, those files have no stored fingerprint → classified as STRUCTURAL → count exceeds the FULL_UPDATE threshold. From that point on, every auto-update triggers FULL_UPDATE, burning hundreds of thousands of tokens that should have been a cheap incremental check.

## Root Cause

Phase 3d step 3 in `hooks/auto-update-prompt.md` instructs the LLM to:
> "Reads the existing fingerprints.json … Merges with existing fingerprints (keep unchanged files as-is) … Writes updated fingerprints.json"

The instruction is correct in intent but too vague — the LLM writes only the freshly computed batch entries to `fingerprints.json` without first loading and preserving the existing entries. The result is a fingerprints.json that shrinks from N to batch-size after every PARTIAL_UPDATE.

**Same failure mode exists in the `/understand` full rebuild (Phase 7, step 2.5)** — if the Node.js fingerprint script is skipped or fails silently, meta.json advances to the new commit hash but fingerprints.json stays at the previous state. Future auto-updates then see all files as "new" (no fingerprint) and escalate to FULL_UPDATE.

## Reproduction

1. Run `/understand` on a project (generates fingerprints.json with N entries)
2. Make a small change to 2–3 files and commit
3. Run auto-update → triggers PARTIAL_UPDATE → writes only 2–3 entries to fingerprints.json (all other N entries lost)
4. Make any further change and commit
5. Run auto-update → all files with lost fingerprints classified as STRUCTURAL → hits FULL_UPDATE threshold

## Evidence (dry-run)

Using `dryrun_ua_fingerprint.py` simulator on a real project (97 files changed, 23 fingerprints stored out of 51 analyzed):

```
SCENARIO: CURRENT (bug) — 23 fingerprints stored
  Source files changed: 97
  NONE      : 0
  COSMETIC  : 0
  STRUCTURAL: 97
  Action    : *** FULL_UPDATE ***

SCENARIO: SIMULATED FIX — 104 fingerprints (all changed files covered)
  Source files changed: 97
  NONE      : 0
  COSMETIC  : 81
[...]

PR #240, opened 2026-05-26
fix(skills/understand): preflight node/pnpm before pipeline starts

simkimsia:
## Summary

Closes the residual silent-failure window from #152: when `node` isn't reachable in the shell `/understand` runs in, Phase 7 step 2 fails silently and Phase 7 step 3 advances `meta.json` anyway — leaving the project with a fresh commit hash recorded but no `fingerprints.json`. The next auto-update sees no stored fingerprint for any file, classifies them all as STRUCTURAL, and escalates every commit to FULL_UPDATE permanently.

Reproduced on plugin v2.7.4 with a user whose nvm/fnm/mise shim was hooked into `~/.zshrc` but not `~/.zshenv`. `node --version` works in her terminal; Claude Code's non-interactive Bash subshell can't see it. `meta.json` + `knowledge-graph.json` get written, no `fingerprints.json`, no error surfaced. Auto-update silently broken from that point onward.

#152's own issue body acknowledges this exact case as a residual gap:

> *"Same failure mode exists in the `/understand` full rebuild (Phase 7, step 2.5) — if the Node.js fingerprint script is skipped or fails silently, meta.json advances to the new commit hash but fingerprints.json stays at the previous state."*

The 2.7.3 fix added the prompt-level "abort Phase 7 if fingerprints fail" instruction. The case above demonstrates that prompt-level abort isn't always honoured by the LLM executor — a deterministic preflight that exits before Phase 7 even runs is the complement.

## Changes

- New `understand-anything-plugin/skills/understand/preflight.sh` — verifies `node` is on PATH, Node major version >= 22, and `pnpm` is on PATH. Each failure exits 1 with an actionable message (e.g., the node-missing message specifically names nvm/fnm/mise/asdf as common triggers and points users at `~/.zshenv` vs `~/.zshrc`).
- `understand-anything-plugin/skills/understand/SKILL.md` — Phase 0 step 1.5 now invokes `bash "\$PLUGIN_ROOT/skills/understand/preflight.sh" || exit 1` after plugin-root resolution and before the conditional `pnpm install` / core build.
- New `tests/skill/understand/test_preflight.test.mjs` — 5 vitest cases covering the happy path, missing node, old node (<22, shim returns v20.11.0), missing pnpm, and broken node binary.

## Why this should be acceptable where #201 wasn't

Aware of #201's rejection grounds — addressing each:

- **"Hard-codes macOS-specific paths"** — no platform-specific paths anywhere. `command -v` and `node --version` only. Cross-platform.
- **"~15 lines of shell preamble in skill files"** — SKILL.md gains 11 lines (1 invocation + 10-line comment). All logic lives in a bundled script — same pattern as `build-fingerprints.mjs` (e7af9ae) and `scan-project.mjs`, both already accepted.
- **"Token/prompt-noise cost for an edge case"** — happy path prints one line of stdout and exits in <100ms. No prompt-side cost.
- **"Fallback masks 'not installed' vs. 'not on PATH'"** — this PR does the opposite. There's no fallback; we just refuse to proceed with an honest error. Directly aligned with the stated maintainer preference.

The qualitative difference from #201: that PR worked around a *visible* error users could already see and fix. This one addresses a *silent* failure mode where users have no idea they're running with a corrupt baseline until weeks later when token spend spikes.

## Test plan

- [x] `pnpm lint` — clean
- [x] `pnpm test` — 201/201 (includes the 5 new preflight cases)
- [x] `pnpm --filter @understand-anything/core test` — 670/670
- [x] `pnpm --filter @understand-anything/skill test` — passes (delegates to root)
- [x] `pnpm build` — all packages build (dashboard chunk-size warning is pre-existing, not introduced by this PR)
- [x] Manual: `env -i PATH=...` invocations confirm each error path surfaces the expected message (missing node, old node, missing pnpm, broken node binary)

## Compatibility

- Backwards compatible. Users with node + pnpm on PATH see one new line of stdout from the preflight, then everything else behaves identically.
- No new dependencies. No CI changes needed (existing `pnpm test` step picks up the new test file via the root `vitest.config.ts`).
- No version bump required (per CONTRIBUTING or maintainer preference).

## Related

- #152 — original auto-update FULL_UPDATE cascade (closed). This PR closes the residual `/understand --full` half explicitly called out in #152's issue body.
- #162 — silent-failure pattern in `extract-structure.mjs` (closed). Same family of "bundled scripts can fail silently and the LLM doesn't notice."

### DIFF
--- a/tests/skill/understand/test_preflight.test.mjs
+++ b/tests/skill/understand/test_preflight.test.mjs
@@ -1,0 +1,187 @@
+import { describe, it, expect, afterEach } from 'vitest';
+import {
+  mkdtempSync,
+  writeFileSync,
+  chmodSync,
+  rmSync,
+} from 'node:fs';
+import { tmpdir } from 'node:os';
+import { join, dirname, resolve } from 'node:path';
+import { spawnSync } from 'node:child_process';
+import { fileURLToPath } from 'node:url';
+
+const __dirname = dirname(fileURLToPath(import.meta.url));
+const SCRIPT = resolve(
+  __dirname,
+  '../../../understand-anything-plugin/skills/understand/preflight.sh',
+);
+
+/**
+ * Resolve the real node binary on PATH so we can put its directory into
+ * tightly-scoped test PATHs (vs. inheriting the parent process's full PATH,
+ * which would let preflight see things we don't want it to in negative tests).
+ */
+function realNodeDir() {
+  const which = spawnSync('which', ['node'], { encoding: 'utf-8' });
+  if (which.status !== 0 || !which.stdout.trim()) {
+    throw new Error('test environment has no `node` on PATH — cannot test preflight.sh');
+  }
+  return dirname(which.stdout.trim());
+}
+
+function realPnpmDir() {
+  const which = spawnSync('which', ['pnpm'], { encoding: 'utf-8' });
+  if (which.status !== 0 || !which.stdout.trim()) {
+    return null; // pnpm may be unavailable in some envs — happy-path test will skip
+  }
+  return dirname(which.stdout.trim());
+}
+
+/**
+ * Resolve bash by absolute path. spawnSync uses the *child env's* PATH to
+ * locate the executable, not the parent's — so tests that strip PATH down to
+ * just a fixture directory would otherwise fail with ENOENT for bash itself.
+ */
+const BASH = (() => {
+  const which = spawnSync('which', ['bash'], { encoding: 'utf-8' });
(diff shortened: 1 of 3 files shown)


simkimsia:
I have also noticed that the demo repo https://github.com/Lum1104/microservices-demo/tree/main/.understand-anything is also missing fingerprints.json

One possible reason is that the demo repo was executed before the commit `e7af9ae — fix(skills/understand): bundle build-fingerprints.mjs and reorder Phase 7` 

Regardless, the demo repo should be re-run after this PR is merged and checked for fingerprints.json

PR #243, opened 2026-05-26
feat(scan): add --exclude CLI flag for /understand (#76)

Ylsssq926:
Addresses sub-task 4 from #76 — letting users narrow the scan scope from the command line without editing .understandignore.

**What it does:**

`/understand --exclude='**/*.test.*' --exclude='docs/'` now excludes matching files from the knowledge graph. Patterns use the same gitignore glob syntax as `.understandignore`, and stack on top of it (applied as the last layer).

**Changes:**

- `ignore-filter.ts`: `createIgnoreFilter` accepts an optional `{ extraExclude }` parameter, appended after all file-based layers
- `scan-project.mjs`: new `parseArgs()` collects `--exclude=<pattern>` flags (repeatable, equals-separated to match `compute-batches.mjs` style). Output JSON includes `appliedExclude` when patterns are active so downstream agents can verify what was applied
- `SKILL.md` + `project-scanner.md`: documented the new flag, its syntax, and limitations (can't override .gitignore since those files are filtered at `git ls-files` enumeration time)
- 6 unit tests for `createIgnoreFilter` with `extraExclude`
- 6 integration tests for `scan-project.mjs` CLI behavior

**Design choices:**

- Only `--exclude` for now. `--include` has limited utility (can't rescue .gitignore-filtered files) and adds semantic complexity — happy to add it as a follow-up if there's demand.
- Unknown `--flag` tokens are silently ignored for forward compatibility with future agent options.
- Empty `--exclude=` values are skipped with no error (defensive, avoids accidental wildcard match).
- `filteredByIgnore` count now covers both `.understandignore` and CLI excludes (same accounting bucket).

All existing tests pass. The 6+6 new tests cover single/multi pattern, stacking with .understandignore, empty value handling, unknown flag tolerance, and backward compat (no flags = identical behavior).

### DIFF
--- a/tests/skill/understand/test_scan_project.test.mjs
+++ b/tests/skill/understand/test_scan_project.test.mjs
@@ -60,11 +60,11 @@ const _runScriptOutputDirs = [];
  * { status, stdout, stderr, output } where `output` is the parsed JSON
  * written by the script (or null on failure).
  */
-function runScript(projectRoot) {
+function runScript(projectRoot, extraArgs = []) {
   const outputDir = mkdtempSync(join(tmpdir(), 'ua-scan-out-'));
   _runScriptOutputDirs.push(outputDir);
   const outputPath = join(outputDir, 'scan-output.json');
-  const result = spawnSync('node', [SCRIPT, projectRoot, outputPath], {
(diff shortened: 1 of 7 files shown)


Lum1104:
We use .understandignore to support this feature, thanks

PR #268, opened 2026-05-26
Resume interrupted full analysis from completed batches

### LINKED ISSUE #149: Checkpoint/Resume Mechanism for Large-Scale Code Analysis Interruptions
When understand --full phase 1 triggers a rate limit and gets interrupted, the progress from the first half is lost. Can support be added to resume from where it left off?

0xjc65eth:
## Summary

Adds an automatic checkpoint/resume plan for full `/understand --full` runs:

- detect valid existing `batch-<index>.json` and `batch-<index>-part-<k>.json` outputs
- write `.understand-anything/intermediate/resume-plan.json` with completed and pending batch indexes
- update the `/understand` flow to dispatch only missing or invalid batches before merging
- cover the planner with tests for empty, completed, multipart, malformed, and stale-index checkpoint files

Fixes #149.

## Why

Large codebases can hit rate limits or other interruptions after expensive file-analysis batches have already been written. Before this change, rerunning a full analysis would treat the run as all-or-nothing and repeat work that already completed.

With this change, rerunning `/understand --full` after an interruption can reuse the valid batch fragments already on disk and continue from the remaining batches.

## Testing

- `node --check understand-anything-plugin/skills/understand/plan-resume-batches.mjs`
- `node --check tests/skill/understand/test_plan_resume_batches.test.mjs`
- manual smoke test: created a temporary `.understand-anything/intermediate` with two batches and one valid `batch-1.json`; verified the planner reported batch 1 completed and batch 2 pending
- `git diff --check`

I could not run the Vitest test locally because `pnpm` hangs while bootstrapping the pinned package-manager version in this environment, even for `pnpm --version`.

### DIFF
--- a/tests/skill/understand/test_plan_resume_batches.test.mjs
+++ b/tests/skill/understand/test_plan_resume_batches.test.mjs
@@ -1,0 +1,108 @@
+import { describe, it, expect, afterEach } from 'vitest';
+import { mkdtempSync, mkdirSync, readFileSync, rmSync, writeFileSync } from 'node:fs';
+import { tmpdir } from 'node:os';
+import { join, dirname, resolve } from 'node:path';
+import { spawnSync } from 'node:child_process';
+import { fileURLToPath } from 'node:url';
+
+const __dirname = dirname(fileURLToPath(import.meta.url));
+const SCRIPT = resolve(__dirname, '../../../understand-anything-plugin/skills/understand/plan-resume-batches.mjs');
+
+function runScript(projectRoot) {
+  return spawnSync('node', [SCRIPT, projectRoot], {
+    encoding: 'utf-8',
+  });
+}
+
+function setupProject() {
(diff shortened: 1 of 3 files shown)


chiel:
Yeah this seems like a must tbh, I tried running `/understand` today on my pet project and I hit the limit at batch 20/100, which means it is basically unusable, since I assume running it again would just result in the same issue.

Seems like a great tool though.

0xjc65eth:
Thanks @chiel, that is exactly the failure mode this PR is trying to address.

The intended behavior after this change is:

- if `/understand --full` completes batches 1..20 and then stops,
- rerunning it should validate the existing batch outputs,
- write a resume plan,
- and continue from the missing/invalid batches instead of starting from zero.

That should make large repos much less all-or-nothing. Happy to adjust the resume-plan shape or validation rules if maintainers prefer a different integration point.

PR #275, opened 2026-05-27
fix: fall back to npm when pnpm is unavailable

### LINKED ISSUE #270: pnpm depedency
We use npm instead of pnpm in our dev environment so when i try to run the understand command it fails with an error.

any alternatives so it can fall back to npm

cat0825:
## Summary
- Fall back to npm when pnpm is unavailable while bootstrapping `/understand`
- Apply the same fallback path for `/understand-dashboard`
- Keep pnpm as the preferred package manager when it exists

## Verification
- `pnpm lint`

Closes #270


cat0825:
Merged current `main` into this branch; it is now conflict-free and GitHub reports `MERGEABLE`.

Only `understand-anything-plugin/skills/understand-dashboard/SKILL.md` conflicted: `main` had restructured the skill so step 4 is the prebuilt-viewer fast path and step 5 the install/build fallback, so I kept that structure and moved the pnpm-with-npm-fallback guard into the new step 5, switching it from the old `<plugin-root>` / `<dashboard-dir>` placeholders to the quoted `"$PLUGIN_ROOT"` / `"$DASHBOARD_DIR"` variables the current wording uses. `understand-anything-plugin/skills/understand/SKILL.md` merged cleanly and keeps the same fallback in the core-build step.

This is a docs/skill-markdown change with no runtime code, so there is no unit-test surface of its own, but the skill snippets are asserted by tests: `npx vitest run tests/skill/understand/test_skill_security_snippets.test.mjs tests/skill/understand/test_skill_graph_freshness.test.mjs tests/install/platform-table-consistency.test.mjs` → 3 files passed, 17 tests passed. That suite is what would have caught the stale placeholders, since it asserts `cd "$PLUGIN_ROOT" && pnpm --filter @understand-anything/core build` and rejects any `<plugin-root>` form.

One thing outside my control: the CI run is sitting at `action_required` (first-time-contributor workflow approval), which is why the merge state shows `UNSTABLE` rather than `CLEAN` — it needs a maintainer to approve the run.

@Lum1104 no rush, but whenever you have a moment this one and #276/#277 are all conflict-free now and ready for a look.

PR #277, opened 2026-05-27
feat: add deterministic cyclomatic complexity

### LINKED ISSUE #273: feat: Add static cyclomatic complexity calculation per function/class node
## Summary

Currently, the `complexity` field on each `GraphNode` is assigned by the LLM as a qualitative 
label (`"simple" | "moderate" | "complex"`). This works well for human-readable summaries, 
but it's not reproducible or queryable as a metric.

It would be useful to also expose a **static, deterministic cyclomatic complexity score** 
calculated from the tree-sitter AST — especially for use cases like prioritizing refactoring 
candidates or integrating with external tooling.

## Proposed Change

Add an optional `cyclomaticComplexity?: number` field to `GraphNode` (and `FunctionInfo` in 
`StructuralAnalysis`), computed by counting decision-point nodes in the tree-sitter AST during 
extraction — before any LLM call.

The existing `traverse()` utility in `base-extractor.ts` already provides everything needed:

```typescript
function computeCyclomaticComplexity(
  node: TreeSitterNode,
  branchTypes: string[]
): number {
  let count = 1;
  traverse(node, (n) => {
    if (branchTypes.includes(n.type)) count++;
  });
  return count;
}
```

Each language extractor would define its own branchTypes array (e.g. if_statement,
for_statement, while_statement, catch_clause, boolean_operator, etc.) based on
tree-sitter grammar node names.

## Design Questions
1. Replace or complement? Should cyclomaticComplexity replace the LLM-assigned complexity label, or live alongside it as an additional field? My preference is to keep both — the label is still useful for quick human reading.
2. Logical operators (&&, ||): Include them in the count (McCabe strict) or exclude (simpler baseline)?
3. Scope: Function-level only, or also aggregate to file-level nodes?

## Motivation
This would make Understand Anything more useful for refactoring workflows — e.g. scoring
candidates by (in-degree × 0.4) + (change_frequency × 0.4) + (cyclomaticComplexity × 0.2)
using the existing calls edges and git metadata.

Happy to contribute a PR if the direction is agreed on.

cat0825:
## Summary
- Add deterministic `cyclomaticComplexity` to graph and structural-analysis types
- Compute complexity from tree-sitter subtrees for TypeScript/JavaScript and Python functions/classes
- Preserve the metric through `extract-structure.mjs` and document it for file-analyzer outputs
- Add regression coverage for metric extraction and output mapping

## Verification
- `pnpm --filter @understand-anything/core test -- --run src/plugins/tree-sitter-plugin.test.ts src/plugins/extractors/__tests__/python-extractor.test.ts`
- `pnpm test -- --run understand-anything-plugin/src/__tests__/extract-structure.test.mjs`
- `pnpm lint`

Closes #273


cat0825:
Merged current `main` into this branch; it is now conflict-free and GitHub reports `MERGEABLE`.

Only `understand-anything-plugin/skills/understand/extract-structure.mjs` conflicted: `main` had moved `buildResult` out into the new `extract-structure-result.mjs` module. I kept that split and re-applied the two `cyclomaticComplexity` pass-throughs (functions and classes) in `extract-structure-result.mjs`, where `buildResult` now lives, rather than reinstating the old in-place copy. Everything else merged cleanly against the current shapes — `computeCyclomaticComplexity` still sits in `base-extractor.ts` next to `traverse`, both `typescript-extractor.ts` and `python-extractor.ts` call it on the current node shapes, and the `schema.ts` / `types.ts` additions landed on the current `GraphNodeSchema` and `StructuralAnalysis` surface. Net diff is back to the original +120/-9 across 9 files, no unrelated changes.

Verified locally against the merged branch:

- `pnpm --filter @understand-anything/core build` (this is the package's `tsc` typecheck) → exit 0, no diagnostics.
- `pnpm --filter @understand-anything/core test` → **46 test files passed, 977 tests passed**, including the new `TreeSitterPlugin > cyclomaticComplexity > computes deterministic complexity for TypeScript functions and classes`.
- `npx vitest run understand-anything-plugin/src/__tests__/extract-structure.test.mjs` → 1 file passed, 11 tests passed, including `buildResult > cyclomaticComplexity > preserves deterministic complexity for extracted functions and classes` — which is the case that actually exercises the relocation above.
- `npx eslint` on the seven touched source files → exit 0, clean.

One thing outside my control: the CI run is sitting at `action_required` (first-time-contributor workflow approval), which is why the merge state shows `UNSTABLE` rather than `CLEAN` — it needs a maintainer to approve the run.

@Lum1104 no rush, but whenever you have a moment this one and #275/#276 are all conflict-free now and ready for a look.

PR #327, opened 2026-05-29
feat(extractor): unwrap AMD define() to expose nested functions, imports, and exports

Gregox01:
Withdrawn by author. Patch needs scrubbing before resubmission.


Gregox01:
Withdrawing — patch needs scrubbing before resubmission. Will re-open from a clean branch.

Issue #293, opened 2026-05-27
Incremental: Phase 7 cleanup deletes scan-result.json, forcing Phase 1 re-run on every subsequent incremental


atlas-architect:
## Summary

The `/understand` skill's Phase 7 cleanup step deletes `intermediate/`, including `scan-result.json`. But Phase 2's incremental path (`compute-batches.mjs --changed-files=...`) requires `scan-result.json` from a prior Phase 1 SCAN to exist as input. So every "incremental" run is forced to re-run Phase 1 — even when the project file inventory hasn't changed structurally.

## Reproducer

1. Run `/understand` (full scan) — completes with Phase 7 cleanup deleting `intermediate/scan-result.json`
2. Commit a small change to one file
3. Run `/understand` again — Phase 2 incremental path calls `compute-batches.mjs --changed-files=...`
4. Script errors: `Error: scan-result.json not found at .understand-anything/intermediate/scan-result.json`
5. The skill has to dispatch a fresh project-scanner subagent (Phase 1 SCAN) to regenerate the inventory

## Economic impact

For a typical Atlas spoke (~280 files), the forced Phase 1 re-run costs ~157k LLM tokens / ~158 sec wall-clock. Multiplied across N spokes for periodic incremental refresh, this is substantial waste — Phase 1's inventory rarely changes structurally between commits.

## Proposed fixes (any one)

**Option A:** Preserve `scan-result.json` between runs — exclude it from Phase 7 cleanup. Other intermediate files (batch-*.json, layers.json, tour.json, etc.) can still be cleaned.

**Option B:** Make `compute-batches.mjs` generate `scan-result.json` on-the-fly when missing, using a lightweight deterministic scan (no LLM dispatch).

**Option C:** Add a fast-path detection: if the changed-files list contains only files whose `fileCategory` is `docs`, `config`, or other non-structural categories, skip Phase 1 entirely and proceed with the existing inventory.

## Context

Discovered during the second incremental scan of an Atlas spoke (CCER, 305-node baseline). Phase 1 was forced to re-run despite only 10 files having changed (1 new component + 1 modified + 3 templates + 5 config bumps). The re-run produced essentially the same inventory as the first scan but cost 157k tokens.

## Related

Sibling to the incremental skill self-inconsistency bug — separate filing.

PR #331, opened 2026-05-30
fix: make phase 7 cleanup reversible

### LINKED ISSUE #301: Phase 7 cleanup uses `rm -rf` which conflicts with destructive-action policies on hardened hosts
## What

`/understand`'s Phase 7 step 4 (`skills/understand/SKILL.md`) cleans up its scratch directories with:

```bash
rm -rf $PROJECT_ROOT/.understand-anything/intermediate
rm -rf $PROJECT_ROOT/.understand-anything/tmp
```

On hosts with destructive-action safety policies (mine: 2-hour freshness window, three-search-angle verification, mandatory audit delegation before any `rm`/`mv -f`/`unlink`/etc.), this trips because the directories were created moments earlier by the same skill. My advisory hook fires; my blocking guard doesn't (low-risk classification); the command goes through but doctrine compliance failed.

No data is lost in practice — those dirs are skill-internal scratch already merged into `knowledge-graph.json`. The conflict is about the *shape* of the action, not the outcome.

## Repro

1. Install a PreToolUse hook on the Claude Code / Codex host that gates `rm -rf` against a freshness window (e.g., refuse to delete files modified within the last 2h without an explicit auditor approval).
2. Run `/understand .` on any repo.
3. Phase 7 step 4 fires `rm -rf .understand-anything/intermediate`; the hook either blocks (forcing user to retry with override) or fires an advisory warning that the user is responsible for resolving.

## Fix candidate

Replace the destructive `rm -rf` with a reversible `mv` to a timestamped trash dir, then opportunistically purge old trash on next run:

```bash
TRASH=$PROJECT_ROOT/.understand-anything/.trash-$(date +%s)
mv $PROJECT_ROOT/.understand-anything/intermediate $TRASH/ 2>/dev/null
mv $PROJECT_ROOT/.understand-anything/tmp $TRASH/ 2>/dev/null

# At Phase 0 of subsequent runs, purge .trash-* dirs older than 7 days:
find $PROJECT_ROOT/.understand-anything/ -maxdepth 1 -type d -name '.trash-*' -mtime +7 -exec rm -rf {} +
```

Benefits:

- `mv` is reversible (file metadata preserved), so destructive-action gates that require a freshness check let it through cleanly.
- The 7-day delayed purge gives the user time to recover from a botched run.
- Same end-state for the user (clean `.understand-anything/` after a few runs) without the doctrine conflict.

## Additional context

- Verified on: plugin v2.7.5, macOS 25.3, zsh, Claude Code.
- First-real-repo run: 283-node / 401-edge knowledge graph on a polyglot Python+Markdown+Shell+JS repo (~69 source files post-`.understandignore`).
- Related issue from skill recipe itself: #152 (fingerprint baseline ordering — the right shape for "safety property must be true before next side effect").

Happy to PR if the `mv` → `.trash/` pattern looks right; would also be a tiny doc nit to add "destructive-action safety" to the failure modes the skill recipe acknowledges.

sjh9714:
## Summary

Fixes #301.

Phase 7 no longer asks the host shell to immediately delete freshly-created `.understand-anything/intermediate` and `tmp` work directories. Instead, it calls a bundled cleanup script that moves those directories into timestamped `.trash-*` directories and only purges trash older than seven days.

## Changes

- Add `cleanup-workdirs.mjs` to move `intermediate/` and `tmp/` into reversible trash and prune expired trash.
- Update `/understand` Phase 7 to call the bundled cleanup script.
- Add regression coverage for moving work dirs, retaining recent trash, purging expired trash, and avoiding the old Phase 7 cleanup command.
- Bump plugin manifest versions to `2.7.6` per the repository versioning note.

## Testing

- `pnpm install`
- `node --check understand-anything-plugin/skills/understand/cleanup-workdirs.mjs`
- `pnpm vitest run tests/skill/understand/test_cleanup_workdirs.test.mjs`
- `pnpm lint`
- `pnpm test`
- `pnpm build`
- `git diff --check`

Note: I used Codex while preparing this change, reviewed the final diff, and ran the listed checks locally.

### DIFF
--- a/tests/skill/understand/test_cleanup_workdirs.test.mjs
+++ b/tests/skill/understand/test_cleanup_workdirs.test.mjs
@@ -1,0 +1,120 @@
+import { describe, it, expect, afterEach } from 'vitest';
+import {
+  existsSync,
+  mkdtempSync,
+  mkdirSync,
+  readFileSync,
+  readdirSync,
+  rmSync,
+  utimesSync,
+  writeFileSync,
+} from 'node:fs';
+import { tmpdir } from 'node:os';
+import { join, resolve, dirname } from 'node:path';
+import { spawnSync } from 'node:child_process';
+import { fileURLToPath } from 'node:url';
+
+const __dirname = dirname(fileURLToPath(import.meta.url));
+const SCRIPT = resolve(
+  __dirname,
+  '../../../understand-anything-plugin/skills/understand/cleanup-workdirs.mjs',
+);
+const SKILL = resolve(
+  __dirname,
+  '../../../understand-anything-plugin/skills/understand/SKILL.md',
+);
+const NOW_MS = Date.UTC(2026, 4, 29, 12, 0, 0);
+
+function runCleanup(projectRoot, args = []) {
+  return spawnSync('node', [SCRIPT, projectRoot, `--now-ms=${NOW_MS}`, ...args], {
+    encoding: 'utf-8',
+  });
+}
+
+describe('cleanup-workdirs.mjs', () => {
+  let root;
+
+  afterEach(() => {
+    if (root) rmSync(root, { recursive: true, force: true });
+    root = undefined;
+  });
+
+  it('moves fresh intermediate and tmp directories into timestamped trash', () => {
+    root = mkdtempSync(join(tmpdir(), 'ua-cleanup-'));
+    const stateDir = join(root, '.understand-anything');
+    mkdirSync(join(stateDir, 'intermediate', 'nested'), { recursive: true });
+    mkdirSync(join(stateDir, 'tmp'), { recursive: true });
+    writeFileSync(join(stateDir, 'intermediate', 'nested', 'graph.json'), '{"ok":true}');
+    writeFileSync(join(stateDir, 'tmp', 'validate.cjs'), 'console.log("ok");');
+
+    const result = runCleanup(root);
+
+    expect(result.status).toBe(0);
+    expect(result.stdout).toMatch(/moved intermediate, tmp/);
+    expect(existsSync(join(stateDir, 'intermediate'))).toBe(false);
+    expect(existsSync(join(stateDir, 'tmp'))).toBe(false);
+
+    const trashDirs = readdirSync(stateDir).filter((name) => name.startsWith('.trash-'));
+    expect(trashDirs).toHaveLength(1);
+    const trashDir = join(stateDir, trashDirs[0]);
+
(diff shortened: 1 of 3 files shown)


sjh9714:
Rebased this onto the latest `main` and resolved the Phase 7 cleanup conflict by keeping the shared cleanup script path while preserving the upstream `scan-result.json` behavior.

Verified locally:
- `pnpm test tests/skill/understand/test_cleanup_workdirs.test.mjs`
- `pnpm lint`
- `pnpm test`
- `pnpm build`
- `git diff --check`

tirth8205:
Reviewed and tested in an isolated worktree. This is a clean, well-scoped fix for #301.

**What I verified (by running `cleanup-workdirs.mjs` directly, not just reading it):**
- Cleanup is genuinely reversible — fresh `intermediate/` (minus `scan-result.json`) and `tmp/` are `renameSync`'d into a timestamped `.trash-*` dir; nothing fresh is ever `rm -rf`'d. Only trash older than the 7-day retention (and the emptied `intermediate/` via `rmdirSync`) is deleted. This is exactly the shape #301 asked for.
- `scan-result.json` is preserved in place for incremental runs (#293).
- Idempotent: re-running with nothing fresh is a clean no-op; trash naming (`ISO-ts-with-ms-<pid>`) is collision-resistant; purge runs before move so a just-created trash dir can't be purged in the same run.
- Edge cases hold: `intermediate/` containing only `scan-result.json` creates no orphan trash dir; `rmdirSync` safely no-ops while `scan-result.json` remains; negative/non-finite `--retention-days` is rejected (exit 1); `--retention-days=0` purges immediately.
- The new test passes 4/4 and both new files lint clean. The script depends only on `node:fs`/`node:path` builtins.

(Note: `test_compute_batches.test.mjs` fails in my worktree, but that's a pre-existing env issue — missing `graphology` dep — and is unrelated to this PR, which doesn't touch that script.)

**Minor, non-blocking:**
1. The PR description says it bumps versions to `2.7.6`, but the diff has no version changes — `main` is already `2.7.6`. Just a stale description line; the five version files are in sync.
2. Trash is now purged in two places: the Phase 0 `find ... -mtime +7` (SKILL.md step 3.1) and the script's own `purgeExpiredTrash` on every Phase 7 run. Harmless overlap, but the Phase 0 purge is now largely redundant — worth collapsing to one mechanism, or keeping Phase 0 explicitly as a safety net for runs that abort before Phase 7.

Correctness and safety look solid. LGTM with the two optional cleanups above.

sjh9714:
Closing this because #387 has already been merged for #301. It implements the same reversible Phase 7 cleanup behavior, so keeping this PR open would duplicate the merged fix.

PR #338, opened 2026-05-31
Windows compatibility: LF endings, UTF-8 stdio, cross-platform tests + CI

chabanade:
## What & why

On Windows the test suite fails in several places. Each one is an
environment/portability issue rather than a logic bug — this PR makes the suite
green on **both Linux and Windows** (and adds Windows to CI so it stays that way).

### Fixes
- **CRLF on checkout** — Git for Windows ships `core.autocrlf=true`, which rewrites
  the repo's LF source to CRLF. CRLF breaks shebang stripping in the `.mjs`
  pipeline scripts when loaded through Vite/esbuild (vitest), so e.g.
  `extract-structure.test.mjs` fails with `SyntaxError: Invalid or unexpected token`.
  → add a `.gitattributes` with `* text=auto eol=lf` so the working tree is LF for
  everyone, regardless of their autocrlf setting.
- **Python stdio encoding** — the Python pipeline scripts print non-ASCII (e.g. the
  `—` em-dash). On Windows the default cp1252 codepage mangles it to `?`/`�`,
  failing `merge-recover-imports`. → force `sys.stdout/stderr` to UTF-8.
- **POSIX vs Windows paths** — `worktree-redirect.test.mjs` compared a Node
  `path.join()` Windows path against a git-bash `pwd -P` POSIX path. → normalise
  both sides through `pwd -P` so the assertion is convention-agnostic.
- **ESM loader limitation** — the `extract-import-map` graceful-failure test injects
  failure via a custom ESM resolve hook (`module.register`). On Windows + Node 24
  that makes relative specifiers resolve under a bare `c:` URL scheme and crashes
  with `ERR_UNSUPPORTED_ESM_URL_SCHEME` — a Node loader limitation in the *test
  harness*, not the product. → `it.skipIf(process.platform === 'win32')` (the
  product's degradation path is unchanged and still covered on Linux/macOS).
- **CI** — add `windows-latest` to the job matrix.

### Testing
Green on Windows (Node 24, pnpm) and Linux: core + dashboard + skill suites pass,
with one documented Windows-only skip noted above.

---
*These came out of using the plugin on Windows. I have a few more improvements
(secret-file ignore defaults, theme-aware export, a French locale, an offline
semantic-search mode) on my fork — happy to send focused follow-up PRs if useful.*

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,20 @@
+# Force LF line endings for all text files.
+#
+# Why: Git for Windows ships with core.autocrlf=true, which rewrites the
+# repo's LF source to CRLF on checkout. CRLF breaks shebang stripping in the
+# .mjs pipeline scripts when loaded through Vite/esbuild (vitest), and mangles
+# shell/Python tooling. Setting eol=lf here overrides autocrlf for every
+# contributor and every Windows user who installs the plugin, so the working
+# tree always matches the committed (LF) blobs.
+* text=auto eol=lf
+
+# Binary assets — never apply EOL conversion.
+*.png   binary
+*.jpg   binary
+*.jpeg  binary
+*.gif   binary
+*.ico   binary
+*.woff  binary
+*.woff2 binary
+*.wasm  binary
+*.pdf   binary
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,7 +5,14 @@ on:
 
 jobs:
   ci:
-    runs-on: ubuntu-latest
+    # Run the suite on Linux AND Windows. Windows-specific breakage (CRLF
+    # checkout, cp1252 stdio, POSIX-vs-Windows paths) used to slip through a
+    # Linux-only CI; this matrix locks in the cross-platform fixes.
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, windows-latest]
+    runs-on: ${{ matrix.os }}
     steps:
       - uses: actions/checkout@v4
 
--- a/tests/skill/understand/test_extract_import_map.test.mjs
+++ b/tests/skill/understand/test_extract_import_map.test.mjs
@@ -1538,7 +1538,13 @@ describe('extract-import-map.mjs — tree-sitter init graceful failure', () => {
     }
   });
 
-  it('emits a Warning: and produces empty importMap entries when tree-sitter init throws', () => {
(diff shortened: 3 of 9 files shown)


tirth8205:
Reviewed by checking out the branch, reading the surrounding code for each changed file, and running the relevant suites on macOS/POSIX. This is a clean, well-scoped cross-platform fix — LGTM.

**What I verified**

- **`.gitattributes` (`* text=auto eol=lf`)** — Ran `git add --renormalize .`: zero files change, confirming the repo is already all-LF, so this is purely protective against Windows `autocrlf` rewriting LF→CRLF (which breaks `.mjs` shebang stripping through esbuild). No existing blobs are clobbered. Binary-asset exclusions are appropriate.
- **Python UTF-8 stdio** — Correct and minimal: `hasattr`-guarded, placed right after `import sys`. It targets the real gap (the `print`/stderr path); the file reads/writes in these scripts already pass explicit `encoding="utf-8"`. All five scripts parse cleanly and the reconfigure pattern works.
- **`worktree-redirect.test.mjs` `bashPath()`** — Sound. `mainRepo` is already `realpathSync`'d, so routing the expected value through the same `pwd -P` the snippet uses is idempotent on POSIX and convention-agnostic on Windows. Passes locally.
- **`extract-import-map` Windows skip** — Correctly scoped and well-documented: `ERR_UNSUPPORTED_ESM_URL_SCHEME` from a `module.register` resolve hook is a Node-on-Windows *harness* limitation, not a product defect; the product degradation path is unchanged and still covered on Linux/macOS. The skipped test runs and passes on macOS.
- **CI matrix** — Valid YAML, `fail-fast: false`, all run steps are portable `pnpm` commands (no bash-only syntax), so they're fine under the Windows runner's PowerShell.

**Tests run (macOS, Node 24, POSIX):** `worktree-redirect` (5) + `extract-import-map` (44) + full skill suite (95 total) — all green. No POSIX regression.

**Minor, out of scope (no change requested):** `extract-domain-context.py` has a few pre-existing `read_text(errors="replace")` calls without an explicit encoding (default cp1252 on Windows) and a `write_text` at L425 without encoding. The write is safe since its `json.dumps` is ASCII-only, and these predate this PR — just flagging for a possible future Windows pass.

Nice incremental work, and the follow-up ideas in the description (secret-file ignores, theme-aware export, French locale, offline semantic search) sound worth focused PRs.

Issue #226, opened 2026-05-25
iOS project only groups files by domain, but no reference relationship edges are generated


nullLuli:
Hi, thanks for building this project — it looks really promising.

I’m trying to analyze an iOS repository that contains a mix of:
- Swift
- Objective-C
- Flutter/Dart

The analysis completes successfully, and files are grouped into business/domain clusters correctly. However, no relationship edges are generated between files (see screenshot below).

The repository is relatively large (approximately 15,282 files), so I’m not sure whether the issue is related to repository size or language support.

In the demo , files are connected with dashed lines that seem to represent references/import/call relationships. In my case, the graph only shows isolated grouped files without any connections.

### Expected behavior

The generated graph should include relationship edges between related files/modules, similar to the demo.

### Actual behavior
- Files are grouped by business/domain
- No dashed connection lines appear
- No file reference relationships are visible

### Questions

1. Is reference-edge generation currently supported for Swift / Objective-C / Dart projects?
2. Are there additional configuration steps required for iOS repositories?
3. Does the relationship extraction depend on a specific parser, indexer, or build artifacts?
4. Is mixed-language support (Swift + OC + Flutter) currently limited?
5. Could this issue be related to repository size? The project contains approximately 15,282 files.

Thanks!
<img width="1714" height="863" alt="Image" src="https://github.com/user-attachments/assets/41d8dee9-fc97-4951-9173-90947f9e86db" />

ggboom99:
是的我也遇到和你一样的问题，面对大型文件项目它只按域名分组文件，生成不了边导致结果上手的学习效果很差。

PR #346, opened 2026-05-31
perf(understand): parallelise file I/O in compute-batches + extract-import-map (#76)

tirth8205:
## Summary

The /understand pipeline reads every code file twice during analysis:

1. **`compute-batches.mjs`** — `extractExports()` reads + parses every code file once to build the cross-batch neighbour map (each batch's connected-symbol set), so the LLM gets accurate cross-batch edge candidates.
2. **`extract-import-map.mjs`** — `loadTsConfigs` / `loadGoModules` / `loadPhpAutoloads` read every tsconfig.json / go.mod / composer.json to drive cross-file import resolution.

Both sites used sequential `readFileSync` loops. On the iOS repo in #226 (~15k files) that effectively serialises disk reads behind a single libuv thread while the rest of the pool sits idle — and #76 reports first-run `/understand` is slow enough that users open issues about it.

## What changed

| File | Change |
|---|---|
| `compute-batches.mjs` | `extractExports` now batches files into `IO_PARALLELISM = 64` slices and issues all `readFile` calls in each slice through `Promise.all`, letting libuv's worker-thread pool overlap disk reads. Tree-sitter parses stay on the main thread (web-tree-sitter is single-threaded WASM); the savings come from pipelining the I/O while parses run. |
| `extract-import-map.mjs` | `loadTsConfigs`, `loadGoModules`, `loadPhpAutoloads` and `buildResolutionContext` switch to async + `Promise.all`. `buildResolutionContext` also runs the three loader passes concurrently since they're independent. A small `readFilesParallel(paths)` helper at the top of the file shares the error-preserving shape across all three. |

## Why behaviour stays identical

- Each loader collects its candidate paths in `files[]` order **before** issuing reads, then iterates the resulting array in that order to emit warnings + populate output maps. So **stderr order and final map contents are byte-identical** to the previous sequential loops.
- `extractExports` collects per-file errors in-place in the `Promise.all` callbacks and emits warnings during the post-read **serial** loop, again in chunk order — so warning text and order match the previous implementation.
- Tree-sitter parsing is unchanged: parses still run serially on the main thread, just with reads pipelined alongside.

## What's NOT in this PR

- `buildFingerprintStore` and `analyzeChanges` in `core/fingerprint.ts` have the same sequential pattern. They're left alone here because they're part of the public `@understand-anything/core` API; making them async would be a breaking change worth its own discussion.
- No change to `scan-project.mjs`: most of its sync I/O is `statSync` (metadata, not content) plus a handful of small `.gitignore` / `.understandignore` reads. The parallelism win is marginal there.

## Verification

- ✅ `pnpm lint` clean
- ✅ `pnpm --filter @understand-anything/core build` clean
- ✅ `pnpm --filter @understand-anything/skill build` clean
- ✅ `pnpm test`: **196/196** — including the two suites that exercise the changed pipelines end-to-end with fixture projects:
  - `test_compute_batches.test.mjs` (19 tests, exports extraction + batching)
  - `test_extract_import_map.test.mjs` (40 tests, all language resolvers)
  - No output diff vs `main`.

## Test plan

- [ ] Reviewer runs `pnpm install && pnpm test` — all suites pass
- [ ] (Optional, harder) Run `/understand --full` on a ~5k-file repo and compare wall time against `main` — the speedup should be visible on first runs, where the I/O cost dominates

Refs #76

### DIFF
--- a/tests/skill/understand/test_extract_import_map.test.mjs
+++ b/tests/skill/understand/test_extract_import_map.test.mjs
@@ -1492,3 +1492,55 @@ describe('extract-import-map.mjs — tree-sitter init graceful failure', () => {
     expect(result.output.stats.totalEdges).toBe(0);
   });
 });
+
+describe('extract-import-map.mjs — deterministic stderr ordering across loaders', () => {
+  let projectRoot;
+
+  afterEach(() => {
+    if (projectRoot) {
+      rmSync(projectRoot, { recursive: true, force: true });
+      projectRoot = null;
+    }
+  });
+
+  // Regression for the parallel-loader stderr-order bug surfaced in
+  // PR #346 review: tsconfig / go.mod / composer.json loaders now run
+  // concurrently, but warnings must still emit in the pre-PR canonical
+  // order (tsconfig → go → php). If the loaders streamed warnings
+  // mid-flight, I/O timing could reorder them — the assertions below
+  // catch that regression.
+  it('emits warnings in canonical order (tsconfig, go, php) regardless of I/O timing', () => {
+    projectRoot = setupTree({
+      'tsconfig.json': '{ "compilerOptions": { "baseUrl": ".", ', // unterminated
+      'composer.json': '{ "autoload": { "psr-4": { "App\\\\": "src/" }, ', // unterminated
+      'src/index.ts': `import { foo } from './foo';\n`,
+      'src/foo.ts': `export const foo = 1;\n`,
+      'src/Http/Controller.php':
+        `<?php\nnamespace App\\Http;\n\nuse App\\Models\\User;\n\nclass Controller { }\n`,
+      'src/Models/User.php':
+        `<?php\nnamespace App\\Models;\nclass User { }\n`,
+    });
+
+    const result = runScript(projectRoot, {
+      projectRoot,
+      files: [
+        { path: 'tsconfig.json', language: 'json', fileCategory: 'config' },
+        { path: 'composer.json', language: 'json', fileCategory: 'config' },
+        { path: 'src/index.ts', language: 'typescript', fileCategory: 'code' },
+        { path: 'src/foo.ts', language: 'typescript', fileCategory: 'code' },
+        { path: 'src/Http/Controller.php', language: 'php', fileCategory: 'code' },
+        { path: 'src/Models/User.php', language: 'php', fileCategory: 'code' },
+      ],
+    });
+
+    expect(result.status).toBe(0);
+    const tsLineIdx = result.stderr.indexOf('tsconfig.json at');
+    const composerLineIdx = result.stderr.indexOf('composer.json at');
+    expect(tsLineIdx).toBeGreaterThanOrEqual(0);
+    expect(composerLineIdx).toBeGreaterThanOrEqual(0);
+    // Canonical order: tsconfig warnings precede composer warnings.
+    // Pre-PR-346 this fell out of sequential loader passes; post-fix it
+    // falls out of buffering + ordered drain in buildResolutionContext.
+    expect(tsLineIdx).toBeLessThan(composerLineIdx);
+  });
+});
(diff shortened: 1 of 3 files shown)


ZebangCheng:
I verified PR #346 on Windows using `TanStack/query` as a larger TypeScript fixture.

Positive checks:

- `extract-import-map` JSON output matched the before-fix output byte-for-byte.
- Full `batches.json` matched before-fix output byte-for-byte.
- Incremental `compute-batches --changed-files=...` output matched before-fix output byte-for-byte.
- `corepack pnpm -r build`: pass.
- `corepack pnpm lint`: pass.
- Dashboard loaded a verification graph successfully: `260` nodes, `500` edges, `10` layers, diff overlay enabled with `11` changed nodes and `28` affected nodes.

However, I found a deterministic stderr-order regression in `extract-import-map.mjs`.

Repro fixture:

- malformed large `tsconfig.json`
- malformed `composer.json`
- one simple `src/index.ts`

Before PR:

```text
tsconfig, composer
```

After PR:

```text
composer, tsconfig
```

Cause: `buildResolutionContext()` now runs `loadTsConfigs`, `loadGoModules`, and `loadPhpAutoloads` concurrently, but each loader still writes warnings directly to stderr. That allows warnings from independent loader families to interleave based on timing. This contradicts the PR claim that stderr order remains byte-identical.

Suggested fix: buffer warnings inside the async loaders and emit them in the previous deterministic order after the concurrent reads/parses complete.

Verdict: request changes before merging. The performance direction is good, but the current revision changes observable log ordering.

tirth8205:
Thanks for catching this — verified the interleave is real (warnings from `loadTsConfigs` / `loadPhpAutoloads` could land in either order under concurrent `Promise.all`). Fixed in 1f8d165:

- Each of `loadTsConfigs` / `loadGoModules` / `loadPhpAutoloads` now collects warnings into a private array instead of `process.stderr.write` inline, and returns `{ data, warnings }`.
- `buildResolutionContext` drains the buffers in the pre-PR canonical order (`tsconfig → go → php`) **after** `Promise.all` resolves, so stderr is byte-identical to the sequential implementation regardless of I/O timing.

Added a regression test (`extract-import-map.mjs — deterministic stderr ordering across loaders`) that fixtures both a malformed `tsconfig.json` and a malformed `composer.json` and asserts the tsconfig warning precedes the composer warning in stderr. Without the fix this test was flaky-by-timing; with the buffered drain it's deterministic.

Verification:
- `pnpm test`: **197/197** (was 196/196 before — the new regression test is the +1)
- `pnpm lint`: clean

Ready for another look when you have a moment.

ZebangCheng:
I re-verified the updated PR #346 on Windows using the same workflow as before, with `TanStack/query` as the larger TypeScript fixture.

Verdict: this revision resolves the stderr ordering regression I previously reported. From this verification, I think the PR is OK to merge.

Checks performed:

- Reproduced the old problem on previous faulty PR head `eea73b656d168ec91754d9436e60f285c36f2e85`.
- Verified the current PR head `1f8d165f86491f5931ab27aa059066fdb0c18111`.
- Confirmed the malformed `tsconfig.json` + malformed `composer.json` fixture now emits warnings in deterministic order:

```text
tsconfig, composer
```

- Confirmed the old faulty head emitted the regressed order:

```text
composer, tsconfig
```

- Confirmed the warning-order fixture's JSON output hash is unchanged between old and current heads.
- Confirmed `extract-import-map` output on `TanStack/query` is byte-identical between old and current heads.
- Confirmed full `compute-batches` output is byte-identical.
- Confirmed incremental `compute-batches --changed-files=...` output is byte-identical.
- `corepack pnpm -r build`: pass.
- `corepack pnpm lint`: pass.
- The new `deterministic stderr ordering across loaders` test passes.
- Dashboard loaded the generated verification graph successfully: 260 nodes, 500 edges, 10 layers, diff overlay enabled with 10 changed nodes and 29 affected nodes.

7987d31e62781244b58efd395583c5fd


Thanks for fixing the warning buffering/order issue.

PR #347, opened 2026-05-31
feat(core): add Kotlin structural analysis via tree-sitter

tirth8205:
## Summary

Wires Kotlin into the existing tree-sitter pipeline so `.kt` and `.kts` files now produce functions, classes, data classes, sealed classes, interfaces, objects, imports, exports, and call-graph edges — matching the behavior of the other language extractors.

`kotlinConfig` already existed as a stub (no `treeSitter` field), which is why Android / JVM / Gradle codebases currently produce no structural edges between Kotlin files. This PR adds the field, the extractor, and tests; the existing plugin loader picks it up unchanged.

## Why `@tree-sitter-grammars/tree-sitter-kotlin`

The standard `tree-sitter-kotlin` npm package only ships native `.node` prebuilds. **`@tree-sitter-grammars/tree-sitter-kotlin@1.1.0`** ships a prebuilt `.wasm` that loads cleanly with this project's `web-tree-sitter@^0.26.6` loader (verified: nodeTypeCount=289, parses `class_declaration` / `function_declaration` as expected). It's published by the official tree-sitter org account, so it's the canonical upstream WASM source.

Same packaging shape PR1 used for Swift, just a different publisher (the repomix WASM bundle does not include Kotlin).

## Notes for reviewers

| Topic | Detail |
|---|---|
| **Visibility differs from Swift** | Kotlin's default visibility is `public`, so the extractor treats *every* declaration with no modifier as exported. Only an explicit `private` opts out. `internal` and `protected` remain exported in the project-graph sense (still resolvable across files / via inheritance). |
| **`class_declaration` is overloaded** | Same node type for `class`, `data class`, `sealed class`, and `interface` (distinguished by keyword child + `modifiers > class_modifier`). The extractor handles all four uniformly. |
| **`object_declaration`** | Kotlin singletons are a separate node type — surfaced as a class-like entry. |
| **Primary-constructor properties** | Parameters declared `val` / `var` in the primary constructor are surfaced as class `properties`; plain `parameter`s without `val/var` are constructor-only and are NOT counted as properties (matches Kotlin semantics). |
| **Imports** | Three forms handled: plain dotted (`import a.b.C` → specifier=`C`), wildcard (`import a.b.*` → specifier=`*`), aliased (`import a.b.C as Foo` → specifier=`Foo`). |

## Verification

- ✅ `pnpm lint` clean
- ✅ `pnpm --filter @understand-anything/core build` clean
- ✅ `pnpm --filter @understand-anything/skill build` clean
- ✅ `pnpm --filter @understand-anything/core test`: **692/692** passing (+22 new Kotlin cases matching the bar set by `go-extractor.test.ts` / `swift-extractor.test.ts`)
- ✅ `pnpm test`: 196/196 (no regressions)

## Test plan

- [ ] Reviewer runs `pnpm install` to pull the new dep
- [ ] `pnpm --filter @understand-anything/core test` passes (692 tests, includes 22 new Kotlin cases)
- [ ] Spot check: run `/understand --full` against an Android / Spring-Kotlin project and confirm relationship edges now appear between `.kt` files

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -57,6 +57,9 @@ importers:
 
   understand-anything-plugin/packages/core:
     dependencies:
+      '@tree-sitter-grammars/tree-sitter-kotlin':
+        specifier: 1.1.0
+        version: 1.1.0
       fuse.js:
         specifier: ^7.1.0
         version: 7.1.0
@@ -1119,6 +1122,14 @@ packages:
     resolution: {integrity: sha512-TBf2sJjYeb28jD2U/OhwdW0bbOsxkWPwQ7SrqGf9sVcoYwZj7rkXljroBO9wKBut9XnmQLXanuDUeqQK0lGg/w==}
     peerDependencies:
       vite: ^5.2.0 || ^6 || ^7
+
+  '@tree-sitter-grammars/tree-sitter-kotlin@1.1.0':
+    resolution: {integrity: sha512-vlVXaxEE8t2kpJgfZpa8XVvxcnKw9AYtRTgy7KWjsDmAsadk06RxAT80IXOgGQnmM9i/orQn1nD84gPNUHu6DQ==}
+    peerDependencies:
+      tree-sitter: ^0.22.4
+    peerDependenciesMeta:
+      tree-sitter:
+        optional: true
 
   '@types/babel__core@7.20.5':
     resolution: {integrity: sha512-qoQprZvz5wQFJwMDqeseRXWv3rqMvhgpbXFfVyWhbx9X47POIA6i/+dXefEmZKoAgOaTdaIgNSMqMIU61yRyzA==}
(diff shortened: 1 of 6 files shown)


wjhssfs (approved):
LGTM. I came at this independently (built the same tree-sitter Kotlin extractor for my own use) and landed on the same design, so I went through this one closely rather than skimming.

Checked out the branch and verified:
- All 22 tests pass, core builds clean.
- Ran it against a large real Kotlin codebase (~560 .kt files). It extracts functions/classes/imports/exports correctly, including on big files (a 593-line resolver came out as 9 functions, 1 class, 59 imports).
- Cross-package import edges resolve through the existing dotted-FQN resolver once the graph is rooted at the repo root. That's the main payoff for me: it turns a previously edge-sparse Kotlin graph (structure only, via the LLM fallback) into a connected one.

One minor, non-blocking note: the grammar pins @tree-sitter-grammars/tree-sitter-kotlin@1.1.0, and npm flags it with a potential CWE-502 deprecation pointing at 1.3.1+. As far as I can tell 1.1.0 is the latest published version of that package, so there's nothing to bump to. Just flagging so the audit warning isn't a surprise later.

Approving.

PR #359, opened 2026-05-31
fix(extract-import-map): apply NodeNext .js→.ts rewrite (#294)

### LINKED ISSUE #294: Import resolver misses NodeNext .js→.ts rewrite — near-edgeless graph for ESM TypeScript projects
## Summary

The bundled import resolver (`skills/understand/extract-import-map.mjs`) does not apply the NodeNext/ESM `.js → .ts` (and `.jsx/.mjs/.cjs → .tsx/.mts/.cts`) extension rewrite. As a result, in any TypeScript project that uses `"moduleResolution": "NodeNext"` (or `Node16`/`Bundler` with explicit extensions) — where source files import their **compiled output specifier** (`import { x } from './config.js'` while the file on disk is `config.ts`) — virtually every project-internal import fails to resolve.

Since unresolved `imports` targets are dropped as dangling during the merge step, the resulting knowledge graph ends up **effectively edgeless**: hundreds of file nodes with almost no `imports` edges between them, which removes exactly the dependency structure the graph is meant to show.

## Environment

- Plugin version: **2.7.5** (`@understand-anything/skill`), installed commit `26edf61856fa476e466bda1814819a266a293c47`
- Node: v22.14.0
- Affected file: `skills/understand/extract-import-map.mjs`

## Why this happens (TypeScript NodeNext convention)

Under `moduleResolution: NodeNext`, TypeScript does **not** rewrite import specifiers during compilation — what you write in the `.ts` source is emitted verbatim. Because Node's ESM loader requires explicit file extensions at runtime (where only `dist/config.js` exists), the TS source must already spell the import with the `.js` extension that will only be correct *after* compilation:

```ts
// src/index.ts  — real, idiomatic NodeNext source
import { resolveBackend } from './llm-backend-selector.js'; // on disk: llm-backend-selector.ts
import { loadConfig }     from './config.js';               // on disk: config.ts
```

There is no `.js` file on disk — only `.ts`. This is the officially recommended style for ESM TypeScript and is the default for most new TS-ESM projects today.

## Root cause

`probeWithExtensions()` (around L377–385) only ever **appends** extensions from `TS_EXT_PROBES` (L367) to the literal specifier:

```js
const TS_EXT_PROBES = [
  '.ts', '.tsx', '.js', '.jsx', '.mjs', '.cjs',
  '/index.ts', '/index.tsx', '/index.js', '/index.jsx',
];

function probeWithExtensions(basePath, fileSet) {
  if (!basePath) return null;
  if (fileSet.has(basePath)) return basePath;          // './config.js' not on disk -> miss
  for (const ext of TS_EXT_PROBES) {
    const candidate = basePath + ext;                  // tries 'config.js.ts', 'config.js.tsx', ... -> all miss
    if (fileSet.has(candidate)) return candidate;
  }
  return null;                                          // dropped
}
```

For a specifier that already carries a JS-ish extension, the resolver needs to **strip** that extension and re-probe the stem against the TS source extensions. That rewrite is missing.

## Reproduction (verified against the shipped resolver)

```js
// repro.mjs
import { resolveTsJsImport } from './skills/understand/extract-import-map.mjs';

// A NodeNext project: sources are .ts, imports carry the .js output extension
const fileSet = new Set(['src/index.ts', 'src/config.ts', 'src/util/helpers.ts']);
const ctx = { fileSet, tsConfigs: new Map(), packageJsons: new Map() };

for (const spec of ['./config.js', './util/helpers.js', './config']) {
  const r = resolveTsJsImport(spec, { path: 'src/index.ts' }, ctx);
  console.log(`  '${spec}'  ->  ${r === null ? 'NULL (DROPPED)' : r}`);
}
```

**Actual output (v2.7.5):**

```
  './config.js'        ->  NULL (DROPPED)
  './util/helpers.js'  ->  NULL (DROPPED)
  './config'           ->  src/config.ts        # extensionless still works
```

**Real-world impact:** running `/understand` on a NodeNext TS codebase (332 `.ts` files), the resolver resolved **1 of 896** internal import relations. Every other import was dropped as dangling, leaving a near-edgeless graph. A deterministic one-off augmentation pass (same probe order + same `fileSet`, only adding the `.js→.ts` rewrite) recovered all 896.

## Proposed patch

Strip a trailing JS-ish extension and re-probe the stem. Minimal, deterministic, and order-preserving — extensionless and existing-`.ts` imports are unaffected:

```diff
 function probeWithExtensions(basePath, fileSet) {
   if (!basePath) return null;
   // Exact match (import already had an extension)
   if (fileSet.has(basePath)) return basePath;
   for (const ext of TS_EXT_PROBES) {
     const candidate = basePath + ext;
     if (fileSet.has(candidate)) return candidate;
   }
+  // NodeNext/ESM: a TS source imports its OWN compiled output specifier, so a
+  // .js/.jsx/.mjs/.cjs import inside a .ts file actually targets a
+  // .ts/.tsx/.mts/.cts source on disk. Strip the JS-ish extension, re-probe.
+  const m = /[.](js|jsx|mjs|cjs)$/.exec(basePath);
+  if (m) {
+    const stem = basePath.slice(0, -m[0].length);
+    const map = { js: ['.ts', '.tsx'], jsx: ['.tsx'], mjs: ['.mts'], cjs: ['.cts'] };
+    for (const ext of map[m[1]]) {
+      if (fileSet.has(stem + ext)) return stem + ext;
+    }
+    return probeWithExtensions(stem, fileSet); // also covers /index.* forms
+  }
   return null;
 }
```

**Output with the patch applied (verified):**

```
  './config.js'        ->  src/config.ts
  './util/helpers.js'  ->  src/util/helpers.ts
  './config'           ->  src/config.ts
```

Note `.mts`/`.cts` are added to the rewrite map for completeness (they are valid NodeNext source extensions emitting `.mjs`/`.cjs`); the `.ts`/`.tsx` mapping is the case that matters for the overwhelming majority of projects.

Happy to open a PR if useful.

tirth8205:
## Summary

Closes #294 — the bundled import resolver was producing near-edgeless knowledge graphs for any modern ESM TypeScript project because it didn't apply the NodeNext / Node16 \`.js → .ts\` rewrite.

## Why this matters

Under \`moduleResolution: NodeNext\` (or \`Node16\` / \`Bundler\` with explicit extensions — the default for new TS-ESM projects since 2023), TypeScript does **not** rewrite import specifiers during compilation:

\`\`\`ts
// src/index.ts — real, idiomatic NodeNext source
import { x } from './config.js';   // on disk: config.ts
\`\`\`

Before this fix, \`probeWithExtensions\` only tried **appending** extensions to the import specifier:

\`\`\`
'./config.js' → not in fileSet
'./config.js.ts', './config.js.tsx', './config.js.js', ... → all miss
→ returns null → edge dropped at merge as dangling
\`\`\`

Net result on the reporter's repro: a graph with hundreds of file nodes and almost no \`imports\` edges between them, silently removing exactly the dependency structure the graph is meant to show.

## Fix

New \`NODENEXT_REWRITES\` table maps each compiled-output extension to the TypeScript source extensions that could have produced it:

| Output ext | Source extensions (priority order) |
|---|---|
| \`.js\` | \`.ts\`, \`.tsx\`, \`.js\`, \`.jsx\` |
| \`.jsx\` | \`.tsx\`, \`.jsx\` |
| \`.mjs\` | \`.mts\`, \`.mjs\`, \`.ts\` |
| \`.cjs\` | \`.cts\`, \`.cjs\`, \`.ts\` |

\`probeWithExtensions\` now applies the rewrite when the import already ends with one of these extensions and no such file exists on disk. The rewrite runs **before** the legacy append-extensions loop — otherwise \`./foo.js\` would generate the nonsense candidate \`foo.js.ts\` and the loop would never reach the actual \`foo.ts\`.

## Disambiguation

If both \`config.ts\` and \`config.js\` exist on disk (rare, but possible during a partial migration), \`import './config.js'\` still resolves to the \`.js\` — that's an exact-disk match and what NodeNext compilation actually does. The rewrite only kicks in when the \`.js\` doesn't exist. A dedicated regression test pins this behavior.

## Tests

6 new tests in \`test_extract_import_map.test.mjs\`:
- The main #294 case (\`.js → .ts\`)
- \`.jsx → .tsx\` and \`.mjs → .mts\` rewrites
- Disambiguation when both \`.ts\` and \`.js\` exist on disk
- Pure-JS projects still work (real \`.js → .js\` imports)
- Historical no-extension probes unaffected
- Missing files still return null (rewrite can't invent targets)

Total: 202 tests passing (was 196).

## Verification

- \`pnpm lint\` ✅
- \`pnpm --filter @understand-anything/core test\` — 670/670 ✅
- \`pnpm test\` — 202/202 ✅
- \`pnpm --filter @understand-anything/core build\` ✅
- \`pnpm --filter @understand-anything/skill build\` ✅

## Versioning

N/A — pure bug fix.

Closes #294

### DIFF
--- a/tests/skill/understand/test_extract_import_map.test.mjs
+++ b/tests/skill/understand/test_extract_import_map.test.mjs
@@ -237,6 +237,150 @@ describe('extract-import-map.mjs — TypeScript / JavaScript resolver', () => {
       'packages/foo/src/y.ts',
     );
   });
+
+  // ── #294: NodeNext / ESM TypeScript `.js → .ts` rewrite ────────────────
+  //
+  // Under `moduleResolution: NodeNext`, TypeScript does NOT rewrite import
+  // specifiers during compilation — what you write in the .ts source is
+  // emitted verbatim. Because Node's ESM loader requires explicit file
+  // extensions at runtime, the TS source must already spell the import with
+  // the `.js` extension that will only be correct AFTER compilation:
+  //
+  //   import { x } from './config.js';   // on disk: config.ts
+  //
+  // Before the fix, every such import resolved to null, leaving ESM-TS
+  // projects with a near-edgeless knowledge graph.
+
+  it('resolves NodeNext .js → .ts relative imports (the main #294 case)', () => {
+    projectRoot = setupTree({
+      'src/index.ts': `import { resolveBackend } from './llm-backend-selector.js';\nimport { loadConfig } from './config.js';\n`,
+      'src/llm-backend-selector.ts': `export function resolveBackend() {}\n`,
+      'src/config.ts': `export function loadConfig() {}\n`,
+    });
+
+    const result = runScript(projectRoot, {
+      projectRoot,
+      files: [
+        { path: 'src/index.ts', language: 'typescript', fileCategory: 'code' },
+        { path: 'src/llm-backend-selector.ts', language: 'typescript', fileCategory: 'code' },
+        { path: 'src/config.ts', language: 'typescript', fileCategory: 'code' },
+      ],
+    });
+
+    expect(result.status).toBe(0);
+    expect(result.output.importMap['src/index.ts']).toEqual([
+      'src/config.ts',
+      'src/llm-backend-selector.ts',
+    ]);
+  });
+
+  it('resolves NodeNext .jsx → .tsx and .mjs → .mts rewrites', () => {
+    projectRoot = setupTree({
+      'src/index.ts': `import Comp from './Comp.jsx';\nimport { fn } from './worker.mjs';\n`,
+      'src/Comp.tsx': `export default function Comp() {}\n`,
+      'src/worker.mts': `export function fn() {}\n`,
+    });
+
+    const result = runScript(projectRoot, {
+      projectRoot,
+      files: [
+        { path: 'src/index.ts', language: 'typescript', fileCategory: 'code' },
+        { path: 'src/Comp.tsx', language: 'typescript', fileCategory: 'code' },
+        { path: 'src/worker.mts', language: 'typescript', fileCategory: 'code' },
+      ],
+    });
+
+    expect(result.status).toBe(0);
+    expect(result.output.importMap['src/index.ts']).toEqual([
+      'src/Comp.tsx',
+      'src/worker.mts',
+    ]);
+  });
+
+  it('resolves to the .js when both .ts and .js exist on disk', () => {
+    // Rare but possible during a partial migration: both `config.ts` and
+    // `config.js` exist. An `import './config.js'` is an exact-disk match
+    // and should resolve to that exact file — the NodeNext rewrite only
+    // kicks in when the .js *doesn't* exist on disk. We assert this to
+    // pin the disambiguation and avoid future regressions where the rewrite
+    // accidentally prefers `.ts` over an existing `.js`.
+    projectRoot = setupTree({
+      'src/index.ts': `import { x } from './config.js';\n`,
+      'src/config.ts': `export const x = 1;\n`,
+      'src/config.js': `export const x = 1;\n`,
+    });
+
+    const result = runScript(projectRoot, {
+      projectRoot,
+      files: [
+        { path: 'src/index.ts', language: 'typescript', fileCategory: 'code' },
+        { path: 'src/config.ts', language: 'typescript', fileCategory: 'code' },
+        { path: 'src/config.js', language: 'javascript', fileCategory: 'code' },
+      ],
+    });
+
+    expect(result.status).toBe(0);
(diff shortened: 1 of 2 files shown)


Lum1104:
Hi Tirth, can you resolve the conflict~

tirth8205:
Conflicts resolved by merging `upstream/main` (143b4e4).

Only one real conflict, in `tests/skill/understand/test_extract_import_map.test.mjs` — both this PR and #214 appended new tests to the same `describe` block. Kept both groups since they cover independent fixes:

- #214 path-alias tests with leading `"./"` (4 tests)
- #294 NodeNext `.js → .ts` rewrite tests (6 tests)

`extract-import-map.mjs` auto-merged cleanly; both fixes are present in the merged script. All 50 import-map tests pass locally (206/206 across the suite).

PR #368, opened 2026-06-03
fix: Go extractor drops variadic parameters

Mubashirrrr:
## What

The Go extractor silently drops **variadic parameters** (`args ...int`) from extracted function signatures.

## Why

`extractParams` only collected identifiers from `parameter_declaration` nodes:

```ts
const declarations = findChildren(paramsNode, "parameter_declaration");
```

tree-sitter-go emits a variadic parameter as a distinct **`variadic_parameter_declaration`** node, not `parameter_declaration`. So the variadic name is never collected:

```go
func Sprintf(format string, args ...interface{}) string  // extracted: ["format"]   (missing "args")
func Sum(nums ...int) int                                 // extracted: []           (missing "nums")
```

Variadic params are extremely common in real Go (printf-style APIs, functional-options constructors, `append`-style helpers), so this corrupts the extracted signatures / structural graph for a lot of ordinary code. Java's extractor already handles its equivalent (`spread_parameter`) separately, confirming this is an oversight.

## Fix

Also search `variadic_parameter_declaration`. A variadic is always the last Go parameter, so appending it after the normal declarations preserves order.

## Test

Added `extracts variadic parameters` to `go-extractor.test.ts`:

```
$ vitest run src/plugins/extractors/__tests__/go-extractor.test.ts
✓ 26 passed   (the new test fails against the previous code)
```

### DIFF
--- a/understand-anything-plugin/packages/core/src/plugins/extractors/__tests__/go-extractor.test.ts
+++ b/understand-anything-plugin/packages/core/src/plugins/extractors/__tests__/go-extractor.test.ts
@@ -101,6 +101,29 @@ func divide(a, b float64) (float64, error) {
       parser.delete();
     });
 
+    it("extracts variadic parameters", () => {
+      const { tree, parser, root } = parse(`package main
+
+func Sprintf(format string, args ...interface{}) string {
+    return ""
+}
+
+func Sum(nums ...int) int {
+    return 0
+}
+`);
+      const result = extractor.extractStructure(root);
+
+      expect(result.functions).toHaveLength(2);
+      expect(result.functions[0].name).toBe("Sprintf");
+      expect(result.functions[0].params).toEqual(["format", "args"]);
+      expect(result.functions[1].name).toBe("Sum");
+      expect(result.functions[1].params).toEqual(["nums"]);
+
+      tree.delete();
+      parser.delete();
+    });
+
     it("reports correct line ranges for multi-line functions", () => {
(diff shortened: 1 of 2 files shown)


tirth8205:
Reviewed and tested. The fix is correct and well-scoped.

**Correctness** — Root cause is accurate: tree-sitter-go emits a variadic param (`args ...T`) as a `variadic_parameter_declaration` node, distinct from `parameter_declaration`, so `extractParams` never collected the variadic name. Searching both node types fixes it, and the identifier-collection loop already works for the variadic node (the name is still an `identifier` child).

**Ordering** — Appending the variadic declaration after the normal ones is correct in all valid Go: a variadic parameter must be the trailing parameter, so order is always preserved, not just for the test case.

**Testing** — Ran `vitest run src/plugins/extractors/__tests__/go-extractor.test.ts`: all 26 pass. Confirmed the new `extracts variadic parameters` test fails against the pre-fix code, so it's a genuine regression test. The test covers both mixed (`Sprintf(format string, args ...interface{})`) and variadic-only (`Sum(nums ...int)`) signatures.

Minor notes (non-blocking):
- The `extractParams` JSDoc / inner-loop comment still reference only `parameter_declaration`; a mention of the variadic node would keep docs in sync.
- Open PR #445 also touches `go-extractor.ts`/`go-extractor.test.ts` but in different regions with no semantic overlap — whichever lands second may need a trivial rebase on the test file.

LGTM.

PR #370, opened 2026-06-03
fix: protobuf map<> fields were silently dropped from message definitions

Mubashirrrr:
## What

The protobuf parser silently drops every `map<K, V>` field from a message's extracted `fields[]`.

## Why

```ts
const fieldRegex = /^\s*(?:repeated\s+|optional\s+|required\s+|map<[^>]+>\s+)?\w+\s+(\w+)\s*=/gm;
```

`map<...>` was placed inside the **qualifier** alternation group, next to `repeated`/`optional`/`required`. But those qualifiers precede a *separate* type token, whereas `map<K, V>` **is** the type. After optionally consuming `map<string, int32> `, the rest of the pattern still demands `<type> <name> =` — but only `name =` remains, so the match fails and the field is omitted.

```proto
message Config {
  string name = 1;
  map<string, int32> scores = 2;     // dropped
  map<string, Endpoint> endpoints = 3;  // dropped
}
// extracted fields: ["name"]   (expected: ["name", "scores", "endpoints"])
```

Map fields are common in real `.proto` files, so message summaries are silently incomplete. (The file's doc comment lists what's unsupported — nested messages, oneof, proto2 extensions — and does *not* list maps; the regex visibly tries to handle them, it's just placed wrong.)

## Fix

Move `map<...>` into the type position, as an alternative to `\w+`:

```ts
/^\s*(?:repeated\s+|optional\s+|required\s+)?(?:map<[^>]+>|\w+)\s+(\w+)\s*=/gm
```

## Test

```
$ vitest run src/__tests__/parsers.test.ts
✓ 62 passed   (the new "extracts map<> fields" test fails against the previous code)
```

### DIFF
--- a/understand-anything-plugin/packages/core/src/__tests__/parsers.test.ts
+++ b/understand-anything-plugin/packages/core/src/__tests__/parsers.test.ts
@@ -373,6 +373,18 @@ describe("ProtobufParser", () => {
     expect(result.definitions![0].fields).toContain("emails");
   });
 
+  it("extracts map<> fields", () => {
+    const content = `message Config {
+  string name = 1;
+  map<string, int32> scores = 2;
+  map<string, Endpoint> endpoints = 3;
+}`;
+    const result = parser.analyzeFile("config.proto", content);
+    expect(result.definitions![0].fields).toContain("name");
+    expect(result.definitions![0].fields).toContain("scores");
+    expect(result.definitions![0].fields).toContain("endpoints");
+  });
+
   it("extracts enum definitions", () => {
     const content = "enum Status {\n  UNKNOWN = 0;\n  ACTIVE = 1;\n  INACTIVE = 2;\n}";
     const result = parser.analyzeFile("status.proto", content);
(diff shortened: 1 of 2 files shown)


tirth8205:
Verified and approving.

Checked out the branch, ran `npx vitest run src/__tests__/parsers.test.ts` after `pnpm install` — 62/62 pass, including the new `extracts map<> fields` test.

The root-cause analysis is accurate. `map<K,V>` was in the qualifier alternation next to `repeated`/`optional`/`required`, but those are prefixes before a *separate* type token whereas `map<K,V>` is itself the type — so after consuming `map<...> ` the pattern still required `<type> <name> =`, the match failed, and the field was dropped. Moving `map<[^>]+>` into the type position (`(?:map<[^>]+>|\w+)`) is the correct structural fix, and the inline comment documents it well.

I also ran the old vs new regex directly on edge cases beyond the test:
- `map<string, Endpoint>` (map of message): old dropped it, new captures it ✓
- `map<int32, MyMsg>`: ✓
- `map<string, foo.Bar>` (dotted value type): ✓
- `map<string,int32>` (no space): ✓
- `repeated`/`optional`/`required` and plain fields still parse ✓

(Nested maps aren't legal protobuf, so there's no missed case there.)

Scope is tight — one regex line plus a regression test. Looks good to merge.

PR #382, opened 2026-06-04
feat: SystemVerilog / Verilog / UVM support — structural graph with composition + TLM

StopJoking97:
## Summary

Adds recognition of SystemVerilog/Verilog (`.sv` `.svh` `.v` `.vh`) and a deterministic HDL knowledge graph. For RTL it captures the module instantiation hierarchy; for UVM it extracts the **component composition tree** (`Type::type_id::create(...)`) and the **TLM data path** (`a.b.connect(c.d)`), so a UVM testbench renders as one tree rooted at the test instead of a disconnected inheritance fan-out.

Two parser backends behind one `HdlAnalysis` shape:

- **tree-sitter-systemverilog** (primary) — accurate AST extraction; ships a prebuilt `.wasm` grammar, loaded exactly like the existing grammars.
- **regex** (zero-dep fallback) — also the `AnalyzerPlugin` for the standard structural pipeline; used when the WASM grammar can't load.

The deterministic assembler (`assembleHdlGraph`) is parser-agnostic — both backends produce an identical graph. Reuses existing edge types: `contains`, `depends_on` (instantiation + composition), `inherits`, `publishes` (TLM).

## Linked issue(s)

None — happy to open a tracking issue if you'd prefer to discuss the approach first.

## How I tested this

- [x] `pnpm lint`
- [x] `pnpm --filter @understand-anything/core test` — **720 passing**, incl. SV parser/assembler unit tests, a tree-sitter↔regex equivalence integration test, and a self-contained 15-file `uvm_demo` fixture whose test asserts (by graph traversal) that the whole testbench is reachable from `dut_base_test` via composition.
- [x] `pnpm test` — the only failures locally are 9 pre-existing, env-specific ones unrelated to this change (worktree path separators / a Python merge script / tree-sitter init on Windows). CI will confirm on Linux.
- [x] Manual smoke test: `node scripts/analyze-hdl.mjs <uvm_project>` → 57 nodes / 55 edges; the UVM layer shows the full `dut_base_test → env → agent → {driver, monitor, sequencer}` + scoreboard tree, plus TLM `monitor→scoreboard` / `driver→sequencer`. The tree-sitter and regex backends produce identical graphs.

## Versioning

- [x] N/A — additive feature; leaving the five manifests for you to bump on merge per CLAUDE.md (avoids version churn/conflicts in the PR).

## Notes for review

This adds a deterministic HDL assembler (`buildHdlGraph` / `buildHdlGraphTreeSitter`) alongside the existing `GraphBuilder`, so the SV/UVM design edges (instantiation, inheritance, UVM composition, TLM) can be produced and asserted **without an LLM run**. The parser sits behind a seam (regex ↔ tree-sitter) and the assembler is parser-agnostic. Happy to adjust the architecture to your taste — e.g. fold the extraction into the existing tree-sitter `LanguageExtractor` pipeline, or drop the regex fallback.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -90,6 +90,9 @@ importers:
       tree-sitter-rust:
         specifier: ^0.24.0
         version: 0.24.0
+      tree-sitter-systemverilog:
+        specifier: ^0.3.1
+        version: 0.3.1
       tree-sitter-typescript:
         specifier: ^0.23.2
         version: 0.23.2
@@ -835,6 +838,10 @@ packages:
     resolution: {integrity: sha512-O8jcjabXaleOG9DQ0+ARXWZBTfnP4WNAqzuiJK7ll44AmxGKv/J2M4TPjxjY3znBCfvBXFzucm1twdyFybFqEA==}
     engines: {node: '>=12'}
 
+  '@isaacs/fs-minipass@4.0.1':
+    resolution: {integrity: sha512-wgm9Ehl2jpeqP3zw/7mo3kRHFp5MEDhqAdwy1fTGkHAwnkGOVsgpvQhL8B5n1qlb01jV3n/bI0ZfZp5lWA1k4w==}
+    engines: {node: '>=18.0.0'}
+
   '@istanbuljs/schema@0.1.3':
     resolution: {integrity: sha512-ZXRY4jNvVgSVQ8DL3LTcakaAtXwTVUxE81hslsyD2AtoXW/wVob10HkOJ1X/pAlcI7D+2YoZKg5do8G/w6RYgA==}
     engines: {node: '>=8'}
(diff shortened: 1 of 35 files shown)


tirth8205:
Reviewed by checking out the branch, running the SV/UVM test files + the `language-registry` test, the full core suite, and lint.

**Verified working:**
- Registration is complete and consistent: `systemverilogConfig` added to `builtinLanguageConfigs` (registry count bumped to 41), `SystemVerilogParser` added to `registerAllParsers` (parser count bumped to 13), both exported from `index.ts`. Extensions `.sv/.svh/.v/.vh` route correctly and the no-duplicate-extension invariant still holds.
- The prebuilt tree-sitter-systemverilog WASM (21MB) resolves and loads in this environment; the tree-sitter parser tests pass (the ignored `tree-sitter-systemverilog` build script is irrelevant since the grammar ships prebuilt).
- Both backends (regex + tree-sitter) are asserted against an identical graph contract via two parallel corpus integration tests — the "parser-agnostic assembler" claim is genuinely verified, not just stated.
- Full core suite: **720 passing**. `pnpm lint`: clean. Matches the PR description.

The UVM composition + TLM extraction is the strongest part — modeling the testbench as a convergent composition tree rooted at the test (via `type_id::create`) and resolving `connect()` chains into `publishes` edges is the right call, and the reachability test makes it concrete.

**Things for maintainers to weigh (design, not correctness bugs):**
1. The SV tree-sitter backend deliberately sits outside the existing `TreeSitterPlugin`/`LanguageExtractor` architecture (the config has no `treeSitter` field; there's no `systemverilog-extractor`). In a real `/understand` run, SV files get the **regex** AnalyzerPlugin; the accurate AST backend is only reachable via `buildHdlGraphTreeSitter`/`analyze-hdl.mjs`. You already flag this and invite direction — that's the main architectural decision here.
2. `assembleHdlGraph` hardcodes `filePath.startsWith("rtl/")` for the RTL/UVM layer split. Correct for the fixture layout, but would dump everything into the UVM layer for projects with RTL under e.g. `src/rtl/` or absolute paths. Minor since it's script-path-only today.
3. `resolveChain` only resolves handles created within the same class, so TLM endpoints wired through `config_db`/external handles silently drop their `publishes` edge. Fine as a documented MVP limitation.

No blocking issues. Solid, well-tested feature; the open question is how it should integrate with the existing pipeline.

PR #403, opened 2026-06-05
fix(merge): recognize batch-existing.json so incremental updates don't drop ~75% of nodes

### LINKED ISSUE #292: Incremental: skill text instructs writing 'batch-existing.json' which merge regex silently drops
## Summary

The `/understand` skill's incremental-update path is silently broken. Following the skill text literally produces a knowledge graph missing ~75% of its nodes, with only a stderr warning to indicate something went wrong.

## Reproducer

1. Run `/understand` on a project (full analysis) — get baseline graph at N nodes
2. Commit one small change (e.g. add a new source file)
3. Run `/understand` again — the skill detects "existing graph + changed files" → incremental update path
4. Skill instructs (literally): "Write the pruned existing nodes/edges as `batch-existing.json` in the intermediate directory"
5. Then run `merge-batch-graphs.py`

**Expected:** merged graph contains pruned existing nodes + freshly-analyzed nodes
**Actual:** merged graph contains ONLY freshly-analyzed nodes; pruned existing is dropped

## Root cause

`packages/core/scripts/merge-batch-graphs.py` regex for input filenames is:

```python
re.compile(r'batch-(\d+)(?:-part-(\d+))?\.json')
```

`batch-existing.json` does not match `\d+` (no digits in `existing`), so the file is silently dropped. The script prints a warning to stderr (`Warning: dropped 1 batch file(s) with unrecognized filenames — files: batch-existing.json`) but exits 0.

## Real-world impact

In a 305-node baseline + 10 changed files, the broken incremental output had **92 nodes** (only the new batches). The 213 nodes from the pruned-existing payload were dropped. Workaround: rename to `batch-0.json` → merge re-runs cleanly → 312 nodes (correct).

## Proposed fixes (pick one)

**Option A (smallest delta — skill text fix only):** Update `skills/understand/SKILL.md` Phase 2 "Incremental update path" step 3 from:
> Write the pruned existing nodes/edges as `batch-existing.json`

To:
> Write the pruned existing nodes/edges as `batch-0.json` (or any unused integer filename matching the merge script's `batch-<N>.json` regex)

**Option B (smallest delta — merge regex fix only):** Extend the regex in `merge-batch-graphs.py` to also accept `batch-existing.json`:

```python
re.compile(r'batch-(\d+|existing)(?:-part-(\d+))?\.json')
```

**Option C (cleanest):** Make the dropped-filename warning a hard error (exit non-zero) so silent breakage becomes loud breakage. Pairs with either A or B.

## Why Option A is probably best

Renames the convention to match the existing regex without code changes. Skill text is easier to update than runtime invariants. Future contributors won't accidentally regress the regex.

## Context

Surfaced during fleet-side dogfooding (Atlas Intelligence — same project that filed PR#231 for the emoji-folder `git ls-files -z` fix). Banked at https://github.com/atlas-intelligence-io/hcr-expense-report/issues/1#issuecomment-4558734051.

atlas-architect:
## Problem

`/understand`'s incremental-update path is silently lossy. During an incremental update the skill writes the pruned existing-graph payload as **`batch-existing.json`** alongside the freshly-analyzed `batch-<N>.json` files, then runs `merge-batch-graphs.py`.

But the merge step buckets files by the `batch-(\d+)` filename match, so `batch-existing.json` (no digits) falls through to the **"unrecognized"** branch and is dropped at load — only a stderr warning, exit 0. Net effect: every carried-over node from the previous scan is lost.

In a 305-node baseline + 10 changed files, the incremental output had **92 nodes** — the 213 unchanged-file nodes were gone. (The known workaround was to manually rename `batch-existing.json` → `batch-0.json`.)

This is the bug tracked in #292.

## Fix

Recognize `existing` as a valid logical batch in `merge-batch-graphs.py`:
- the filename match becomes `batch-(\d+|existing)(?:-part-(\d+))?\.json`
- `batch-existing.json` is bucketed/sorted as logical index `-1` so it loads **before** the freshly-analyzed numbered batches

So `batch-existing.json` is now loaded and merged like any other batch instead of being dropped. Genuinely-malformed filenames (e.g. `batch-fused-8-13.json`) still hit the existing "unrecognized → warn" path, unchanged.

## Test

Adds `TestIncrementalExistingBatch` to `tests/skill/understand/test_merge_batch_graphs.py` — runs the script end-to-end on a temp intermediate dir containing `batch-existing.json` + `batch-0.json` and asserts the existing nodes survive the merge (and aren't reported as a dropped/unrecognized filename). Full suite green locally (71 tests).

Surfaced during fleet-side dogfooding (Atlas Intelligence — same project that filed #292 and the emoji-folder `git ls-files -z` fix in #231).

Fixes #292

### DIFF
--- a/tests/skill/understand/test_merge_batch_graphs.py
+++ b/tests/skill/understand/test_merge_batch_graphs.py
@@ -1183,5 +1183,89 @@ class TestUnrecognizedBatchFilename(unittest.TestCase):
         self.assertNotIn("file:src/y.ts", node_ids)
 
 
+# ── Incremental update: batch-existing.json (#292) ────────────────────────
+
+
+class TestIncrementalExistingBatch(unittest.TestCase):
+    """Regression test for #292.
+
+    During an incremental update the skill writes the pruned existing-graph
+    payload as `batch-existing.json` alongside freshly-analyzed `batch-<N>.json`
+    files. Before the fix, `batch-existing.json` failed the `\\d+`-only filename
+    match -> was bucketed as "unrecognized" -> silently dropped at load, losing
+    ~75% of nodes on every incremental run. It must now be recognized, loaded,
+    and merged like any other batch.
+    """
+
+    def setUp(self) -> None:
+        import tempfile
+        self.tmp = Path(tempfile.mkdtemp(prefix="ua-mbg-existing-"))
+        self.intermediate = self.tmp / ".understand-anything" / "intermediate"
+        self.intermediate.mkdir(parents=True, exist_ok=True)
+
+    def tearDown(self) -> None:
+        import shutil
+        shutil.rmtree(self.tmp, ignore_errors=True)
+
+    def _write_batch(self, name: str, nodes: list, edges: list) -> None:
+        import json as _j
+        (self.intermediate / name).write_text(
+            _j.dumps({"nodes": nodes, "edges": edges}),
+            encoding="utf-8",
+        )
+
+    def _run_merge(self) -> tuple[int, str, dict]:
+        import subprocess
+        import json as _j
+        result = subprocess.run(
+            ["python3", str(_MODULE_PATH), str(self.tmp)],
+            capture_output=True, text=True,
+        )
+        out_path = self.intermediate / "assembled-graph.json"
+        assembled = _j.loads(out_path.read_text()) if out_path.exists() else {}
+        return result.returncode, result.stderr, assembled
+
+    def test_batch_existing_is_merged_not_dropped(self) -> None:
+        # batch-existing.json = the unchanged-file nodes carried over from the
+        # previous full scan; batch-0.json = the freshly-analyzed changed files.
+        self._write_batch(
+            "batch-existing.json",
+            [_file_node("src/unchanged-a.ts"), _file_node("src/unchanged-b.ts")],
+            [],
+        )
+        self._write_batch(
+            "batch-0.json",
+            [_file_node("src/changed-c.ts")],
+            [],
+        )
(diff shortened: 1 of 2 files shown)


atlas-architect:
Friendly nudge 🙂 — this PR (teach the merge regex to recognize `batch-existing.json`, fixing the ~75% data loss in incremental scans flagged in #292) has been open since June 5, which I think was right around the Lum1104 → Egonex-AI repo transfer — so it may have quietly slipped through. It's a small, self-contained fix. Happy to rebase or adjust anything if that helps it land. Thanks for taking a look! 🙏

atlas-architect:
Status update from our side, in the interest of keeping the thread honest rather than nudging again.

This PR has been open 50 days without a maintainer response (the repo moved Lum1104 → Egonex-AI shortly after it was filed, so I assume attention moved with it — no complaint intended). The underlying bug blocks our incremental-scan workflow entirely, so **we've vendored this exact patch into our local install** and moved on.

**The PR stays open on purpose.** The fix is small, self-contained, and comes with a regression test; if it's useful to the project it's yours whenever a maintainer has a moment. Nothing here needs a reply from us.

For anyone finding this later with the same symptom — quiet ~75% node loss on incremental updates, only a stderr warning to show for it — the two-hunk change in `merge-batch-graphs.py` is the whole fix: teach the batch-file regex to accept `existing` alongside `\d+`, and sort that bucket first so the carried-over graph loads before the freshly-analyzed batches. Verified on a 3-existing + 1-changed fixture: 4/4 nodes merged, previously 1/4.

Happy to rebase if it's ever picked up.

PR #411, opened 2026-06-07
fix(dashboard): edges attached to containers are never rendered (missing Handles) + portal edge re-routing on expand

maximbelyakov:
## Problem

`ContainerNode` has no React Flow `Handle` components. React Flow silently skips any edge whose endpoint has no handle, so in the layer-detail view:

- container → portal dashed edges never render;
- aggregated container → container edges never render.

Net effect: any layer whose files are grouped into containers looks completely disconnected, even though the graph data has the edges. Single ungrouped files DO render their portal edges, which makes the inconsistency look like bad graph data.

**Repro:** any graph where a layer has ≥1 container with cross-layer edges; open the layer — portals float with no incident edges.

## Fix (commit 1)

- Add invisible target (top) / source (bottom) handles to `ContainerNode`.
- Call `useUpdateNodeInternals` when the container resizes on expand/collapse, so edge anchors follow the new bounds instead of pointing at the stale pre-expansion geometry.

## Follow-up (commit 2)

Stage 1 comments said re-routing portal edges to actual file ids on container expand was deferred ("Task 12 will re-route…") — this implements it: when a container is expanded, its single container→portal edge is replaced with per-file edges (e.g. `service.ts → "REST API" portal`), via a `portalCrossFiles` map computed in Stage 1.

Portal cards now also list the names of the external files behind the aggregated connection count (up to 6, then "+N more") via a new `findExternalNeighborFiles()` helper — before, "REST API — 3 connections" gave no clue WHICH files the layer talks to.

Found while building a graph for a real Java + TypeScript codebase. All 43 dashboard unit tests pass.


tirth8205:
Reviewed and tested on a checkout of this branch. Solid, correctly-scoped fix.

**Correctness — confirmed.** The root cause is accurate: `ContainerNode` rendered no React Flow `Handle`, and React Flow v12 silently skips any edge whose endpoint node has no handle, so `container→portal` and aggregated `container→container` edges never drew. The fix adds invisible Top (target) / Bottom (source) handles matching the convention already used by `PortalNode`/`CustomNode`, so a `container.source(Bottom) → portal.target(Top)` edge can now anchor. `useUpdateNodeInternals` keyed on `width/height/isExpanded` is the documented pattern for re-measuring handles on resize, with a complete deps array.

**Re-routing on expand — verified consistent.** `portalEdgesFinal` only rewrites edges for expanded containers, matches files back via `nodeToContainer`, dedupes, and falls back to the aggregated edge when nothing is emitted. `portalCrossFiles` is pre-filtered by `filteredNodeIds`, and expanded containers do render their child file nodes (`expandedChildNodes`), so the re-routed `source: f` targets a node that actually exists. `findExternalNeighborFiles` is a correct mirror of `findCrossLayerFileNodes` (adds the target-layer node instead of the active one) and is used only for the portal card labels — no mismatch with the edge logic.

**No regression to normal edges.** Collapsed-state and ungrouped-file edges are untouched; selection styling still gates on `strokeDasharray` so portal edges keep their dashed style.

**Tested:** `tsc --noEmit` clean, `eslint` clean on all four changed files, and `vitest run` → 42 passing util tests (PR says 43 — immaterial; these are pure-util tests and don't exercise the rendering path).

Minor:
- `findExternalNeighborFiles` and the re-routing logic have no unit tests. The helper is a pure function and cheap to cover; a couple of cases would pin the active-vs-target-layer direction. (The handle/rendering fix itself isn't unit-testable without a React Flow harness, so manual verification there is fair.)

Nit (pre-existing, not introduced here): `edgeAggregation.ts` line 183 uses a literal NUL byte as a key separator, which makes git treat the file as binary — that's why `gh pr diff` shows "Binary files differ" and hides the new `findExternalNeighborFiles` from the diff. Worth a future cleanup so diffs stay readable, but out of scope for this PR.

LGTM.

PR #415, opened 2026-06-08
feat(core): add Dart language support

junddao:
## Summary

Adds Dart to the tree-sitter pipeline so Flutter/Dart codebases get the same structural analysis (functions, classes, imports/exports) other languages already enjoy.

### What's in this PR

- New `dartConfig` registered in the language registry (`.dart` ext, Dart-flavored concepts and file patterns).
- New `DartExtractor` mapping tree-sitter-dart's grammar to the shared `StructuralAnalysis` shape:
  - top-level `function_signature` / getter / setter → functions
  - `class_definition` / `mixin_declaration` / `extension_declaration` / `enum_declaration` → classes (methods + properties + enum constants)
  - `library_import` / `library_export` (wrapped in `import_or_export`) → imports, with the `_`-prefix Dart visibility convention for exports
- Adds `tree-sitter-dart@^1.0.0` as a dependency. Its npm-published WASM ships an outdated dylink format that fails to load against `web-tree-sitter` 0.26.x, so this PR also bundles a freshly rebuilt WASM at `understand-anything-plugin/packages/core/wasm-grammars/tree-sitter-dart.wasm` and teaches `TreeSitterPlugin` to fall back to bundled grammars when the npm copy is missing or stale. The fallback mechanism is generic — any future grammar with a similar issue can drop a WASM into `wasm-grammars/` and benefit from it.
- Tests: adds `DartExtractor` unit tests covering imports, top-level functions, classes/mixins/enums, and the `_`-prefix visibility rule; updates the language-registry built-in count from 40 → 41.

### Notes

- `extractCallGraph` returns `[]` for Dart on purpose. tree-sitter-dart does not expose a single `call_expression` node — calls are encoded via chained `identifier` + `argument_part` selectors, which require nontrivial heuristics to disambiguate from member access. Semantic call relationships are still surfaced by the LLM-side analysis. Happy to follow up with a separate PR if maintainers want a structural call graph.
- Anonymous extensions (`extension on Foo { ... }`) get a synthetic name (`_AnonymousExtension_<line>`) so they don't collide; they are not added to `exports`.

## Linked issue(s)

N/A

## How I tested this

- [x] `pnpm lint`
- [x] `pnpm --filter @understand-anything/core test` — 678/678 passing (added 8 new DartExtractor unit tests)
- [x] `pnpm test` — 201/201 passing
- [x] Verified the WASM fallback path: with the stale npm WASM in place (which does not load), `TreeSitterPlugin` transparently uses the bundled `wasm-grammars/tree-sitter-dart.wasm` and Dart parsing succeeds.

## Versioning

- [x] Version bumped in all five manifests (2.7.6 → 2.8.0)


thejesh23:
@junddao  I think this PR can be closed as Dart support is added through #435 
cc: @Lum1104

miro-park-cupix (approved):

miro-park-cupix:
Closing as Dart support has been added via #435. Thanks @thejesh23 for the heads up!

Issue #406, opened 2026-06-06
question: Invalid knowledge graph: Missing or invalid project metadata


yearyeardiff:
### Your question

mac操作系统，node:v24.13.0。代码版本：2026-06-06 15点左右拉取的main分支最新代码

<img width="2944" height="1808" alt="Image" src="https://github.com/user-attachments/assets/2a7fd472-e528-49f6-9eec-a17d1f6ffc93" />

项目生成的文件：

<img width="826" height="734" alt="Image" src="https://github.com/user-attachments/assets/9cea63b3-bfd1-4113-97fa-bd9846f6bb13" />

帮忙看看是什么原因，好几天没有解决了

whoami-I:
same problem with me.
html show :"Invalid knowledge graph: Missing or invalid project metadata
"
don't show any content

wusy11122132:
解决了吗，感觉缺失蛮多的

PR #418, opened 2026-06-09
fix: check res.ok before parsing knowledge-graph.json in dashboard

### LINKED ISSUE #288: Invalid knowledge graph: Missing or invalid project metadata
I am getting "Invalid knowledge graph: Missing or invalid project metadata" in a Windows 11 env. I tried it on my own code then I did it on the Understand-Anything codes, same issue.

There seems to be unable to find the following files:
- meta.json
- config.json
- knowledge-graph.json
- diff-overlay.json
- domain-graph.json.

It is unclear which file and where it was looking.

Ideally, we can see a log to figure out the value of ENV variables required to map to the physical folder.

Image
Image
Image
Image

eldar702:
## Summary
Closes #288. The dashboard's `knowledge-graph.json` fetch in `App.tsx` did not check `res.ok` before calling `res.json()` — the four sibling fetches in the same effect (`meta.json`, `config.json`, `diff-overlay.json`, `domain-graph.json`) all do. So when the graph file is missing (e.g. before `/understand` has run → HTTP 404), the 404 body was parsed and run through `validateGraph`, surfacing the misleading **"Invalid knowledge graph: Missing or invalid project metadata"** instead of a clear HTTP error.

Thanks to @ZebangCheng for diagnosing the exact root cause and file in the issue.

## Root cause
`packages/dashboard/src/App.tsx` (knowledge-graph fetch) used `.then((res) => res.json())` with no `res.ok` guard, unlike the sibling fetches.

## Fix
Extracted the fetch+validate step into a small, testable helper `utils/fetchAndValidateGraph.ts` that guards on `res.ok` and returns a discriminated result (`loaded` / `http-error` / `validation-error` / `parse-error` / `network-error`). `App.tsx` now surfaces a clear HTTP/network error for a missing graph instead of a schema-validation message; the valid-graph and invalid-JSON paths are unchanged. (The helper extraction lets this path be unit-tested. `packages/dashboard` already has a vitest harness — what was missing was coverage for this fetch/validate path, which this PR adds.)

## Testing
Added `utils/__tests__/fetchAndValidateGraph.test.ts`: a mocked `fetch` returning `{ok:false, status:404}` now surfaces an HTTP-status error (not the schema message); 200-with-invalid-JSON still yields a validation error; network rejection is handled. RED before the fix, GREEN after. Follow-up (per review): a 200 whose body fails to parse is now reported as `parse-error` (distinct from a transport `network-error`), with a dedicated regression test. Full dashboard suite: 47/47; `tsc -b` clean.

---
*AI-assisted contribution — drafted with Claude and verified against current `main`.*

### DIFF
--- a/understand-anything-plugin/packages/dashboard/src/App.tsx
+++ b/understand-anything-plugin/packages/dashboard/src/App.tsx
@@ -147,10 +147,15 @@ function Dashboard({ accessToken }: { accessToken: string }) {
               console.error(`[graph] dropped: ${issue.message}`);
             }
           }
-        } else if (result.status === "http-error" || result.status === "network-error") {
-          // Guard on res.ok (issue #288): surface a clear HTTP/network error
-          // instead of letting a 404 body fall through to graph validation,
-          // which produced a misleading "Missing or invalid project metadata".
+        } else if (
+          result.status === "http-error" ||
+          result.status === "network-error" ||
+          result.status === "parse-error"
+        ) {
+          // Guard on res.ok (issue #288): surface a clear HTTP/network/parse
+          // error instead of letting a 404 body fall through to graph
+          // validation, which produced a misleading "Missing or invalid
+          // project metadata".
           console.error("Failed to load knowledge graph:", result.error);
           setLoadError(`Failed to load knowledge graph: ${result.error}`);
         } else {
--- a/understand-anything-plugin/packages/dashboard/src/App.tsx
+++ b/understand-anything-plugin/packages/dashboard/src/App.tsx
@@ -1,6 +1,7 @@
 import { useEffect, useState, useMemo, useCallback, lazy, Suspense } from "react";
 import { validateGraph } from "@understand-anything/core/schema";
 import type { GraphIssue } from "@understand-anything/core/schema";
+import { fetchAndValidateGraph } from "./utils/fetchAndValidateGraph";
 import { useDashboardStore } from "./store";
 import GraphView from "./components/GraphView";
 import DomainGraphView from "./components/DomainGraphView";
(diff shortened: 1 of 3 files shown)


tirth8205:
Reviewed and tested. The fix is correct and well-scoped.

**Root cause is accurately diagnosed.** The knowledge-graph fetch called `res.json()` with no `res.ok` guard, so a 404 error body (an object) flowed into `validateGraph` and hit the project-metadata branch in `core/src/schema.ts`, surfacing the misleading "Missing or invalid project metadata". I confirmed that code path. Extracting `fetchAndValidateGraph` with an injectable `fetch` and a discriminated result (`loaded` / `http-error` / `validation-error` / `network-error`) is a clean way to make this testable, and it now matches the four sibling fetches that already guard on `res.ok`.

**Behavior is preserved:** `isKnowledge` still reads `kind` from the raw `data` exactly as the old code did, and the issue-iteration and validation-error branches are unchanged.

**Verification:**
- `npx vitest run src/utils/__tests__/fetchAndValidateGraph.test.ts` → 4/4 pass
- Full dashboard suite → 46/46 pass
- `tsc -b` → clean

Two minor, non-blocking notes:
- A `res.json()` parse failure on a 200 is labeled `network-error` — it's really a body-parse failure. Harmless (surfaced as "Failed to load knowledge graph: ..."), but `parse-error` would be more precise.
- The PR description says `packages/dashboard` "had no test harness" — it already has a `test` script, the vitest devDependency, and several existing test files. What was missing was coverage for this path, not the harness. Description-only.

Solid fix; happy to approve.

eldar702:
Thanks for the careful review and for confirming the root-cause path, @tirth8205 — much appreciated. I've addressed both notes:

1. **`parse-error` precision** — you're right that a `res.json()` failure on a 2xx is a body-parse error, not a transport failure. Pushed `6e214c3`: added a dedicated `parse-error` status to `FetchGraphResult`, relabeled the `res.json()` catch, routed it through the load-error branch in `App.tsx`, and added a regression test (a 200 with an unparseable body → `parse-error`, asserting the message is preserved).
2. **Description accuracy** — fixed. The dashboard already has a vitest harness; what was missing was coverage for this fetch/validate path. Reworded the PR body accordingly and updated the discriminated-result list.

Verification after the follow-up: full dashboard suite **47/47**, `tsc -b` clean. Thanks again!

eldar702:
Closing — this got beaten to merge by #577 (dzivkovi), which fixes the exact same root cause (both #288 and #406: a served-but-empty/error response for `knowledge-graph.json` falling through to `validateGraph` and surfacing the misleading "Missing or invalid project metadata" instead of the real HTTP/network error) and is already merged on `main`.

Thanks again @tirth8205 for the review here — much appreciated, even though this particular branch didn't end up landing.

PR #348, opened 2026-05-31
feat(core): add Dart structural analysis via tree-sitter (#226)

tirth8205:
## Summary

Completes the **Flutter portion of #226** — `.dart` files now produce functions, classes, mixins, extensions, enums, imports, exports, and call-graph edges. Same pattern PR1 used for Swift, just a different grammar shape.

This is a sibling PR to the Swift extractor; together they cover the iOS + Flutter combination called out in #226.

## What's covered

- `function_signature` (top-level + class methods) with params and return type
- `class_definition` — fields, constructors (regular / const / factory), and methods are extracted
- `mixin_declaration`, `extension_declaration`, `enum_declaration` are all surfaced as class-like entries
- `import_or_export` handles `dart:` URIs, `package:` URIs, `as` aliases, and `show` combinator identifiers
- **Privacy rule**: Dart has no `private` keyword — names starting with `_` are library-private. The extractor mirrors this: any non-underscored top-level name is exported

## Grammar wrinkles worth flagging for review

| Wrinkle | How it's handled |
|---|---|
| `function_signature` and `function_body` are **siblings**, not nested | Call-graph walker pushes the function name when entering a `function_body` and resolves the name from the preceding sibling |
| Calls are decomposed: callee → `selector > argument_part` | `extractCalleeForArgumentPart` handles both `foo()` (sibling identifier) and `x.foo()` (prior `selector > unconditional_assignable_selector > identifier`) |
| **web-tree-sitter gotcha**: `parent.child(i)` returns a fresh wrapper, so `===` fails | We compare via `.id` (a tree-unique node identity preserved across wrappers). Documented inline since this is non-obvious and bit me during dev |

## Files

| Area | File | Status |
|---|---|---|
| Config | `languages/configs/dart.ts` | New |
| Index | `languages/configs/index.ts` | + `dartConfig` |
| Extractor | `plugins/extractors/dart-extractor.ts` | New |
| Registration | `plugins/extractors/index.ts` | + `DartExtractor` |
| Tests | `plugins/extractors/__tests__/dart-extractor.test.ts` | New (21 cases) |
| Test bump | `__tests__/language-registry.test.ts` | Expected count 40 → 41 for the new builtin config |
| Dep | `core/package.json` | + `@repomix/tree-sitter-wasms@0.1.17` (same bundle as Swift — also ships `tree-sitter-dart.wasm`) |

## Verification

- ✅ `pnpm lint` clean
- ✅ `pnpm --filter @understand-anything/core build` clean
[...]

PR #435, opened 2026-06-13
feat(core): add Dart language support via workspace-vendored WASM

thejesh23:
## Summary

Adds Dart to the tree-sitter pipeline so Flutter / Dart codebases get the same structural analysis (functions, classes with fields/constructors, mixins, extensions, enums, imports, exports) and call-graph edges that other languages already produce. Ships the freshly-built `dylink.0`-format wasm as a workspace-internal package so no upstream-republish or third-party-bundle dependency is required.

## Linked issue(s)

Refs #226 — Dart is the third leg of the iOS-repo case (Swift + Objective-C + Flutter/Dart) where files currently group by domain but produce no reference edges.

## Related PRs (this is the third Dart attempt — flagging explicitly)

There are already two open Dart PRs against `main`. This PR is opened to give maintainers a third angle with explicit trade-off comparison; happy to close it in favour of either of the others if that's the call.

- **#348** (`tirth8205`, May 31, 957 additions) — uses `@repomix/tree-sitter-wasms@0.1.17` for the grammar (a community bundle). Implements the full call graph and references #226 explicitly. Closest sibling to this PR.
- **#415** (`junddao`, Jun 8, 822 additions) — depends on the broken upstream `tree-sitter-dart@1.0.0`, bundles a freshly-rebuilt wasm under `core/wasm-grammars/`, and adds a generic loader fallback to `TreeSitterPlugin` for stale-npm-wasm cases. Skips the structural call graph (returns `[]`).

### What this PR does differently

- **No third-party grammar bundle dep, no loader changes.** Ships the freshly-built wasm as a workspace-internal package `@understand-anything/tree-sitter-dart-wasm/` consumed via `workspace:*`. The existing `TreeSitterPlugin.init()` resolves it through `require.resolve()` unchanged. If `amaanq/tree-sitter-dart` ever publishes a refreshed `dylink.0` wasm, the workspace package is a single-commit delete + dependency-flip.
- **Implements the call graph.** Uses `startIndex` to compare sibling-identity (web-tree-sitter's `.child(i)` returns a fresh JS wrapper per call, so `===` always fails — confirmed via probe; PR #348 hit the same gotcha and uses `.id` for the same purpose).
- **Documents the AST surprises inline.** `function_signature` / `function_body` are siblings not parent/child; abstract methods land at `class_body > declaration > function_signature` (no `method_signature` wrapper); anonymous extensions surface as ``\`on <TargetType>\``` so the graph builder doesn't drop them; `hide` combinators are skipped (not added as specifiers).
- **22 vitest cases** driven against the real wasm — coverage matrix is in the spec doc (`docs/superpowers/specs/2026-06-13-dart-support-design.md`).
- Includes the design spec + bite-sized TDD implementation plan under `docs/superpowers/` for review traceability.

### Honest trade-offs vs. the alternatives

- Adds a ~745 KB vendored binary to git (comparable to the wasms already pulled in at install time for Rust/Go/TS — those just aren't in-repo). Mitigated by the workspace package being trivially deletable when upstream republishes.
- `extractParams` currently surfaces only required positional parameters; named (`{...}`) and optional positional (`[...]`) parameters are wrapped in container nodes and dropped. Documented inline in the helper's doc comment.
- Getters / setters (`int get value => count;`) are not surfaced as methods — they appear as `getter_signature`, which the extractor doesn't currently route. Documented with an explicit test asserting the current (limited) behaviour so the gap can't silently regress.
- Records and pattern matching are out of scope (function-local; no project-graph impact).

## How I tested this

- [x] `pnpm lint` — clean
- [x] `pnpm --filter @understand-anything/core build` — tsc clean
- [x] `pnpm --filter @understand-anything/core test` — 720/720 (+ 22 new Dart cases, + 1 count-assertion update from 40 → 41 in `language-registry.test.ts`)
- [x] `pnpm --filter @understand-anything/skill build` — clean
- [x] `pnpm test` — 201/201 (one pre-existing failure in `tests/skill/merge-recover-imports.test.mjs` exists on `origin/main` unrelated to this branch)
- [x] Manual smoke against `TreeSitterPlugin` with the sample Flutter snippet (`Counter` class + `main()` calling `increment()`): `analyzeFile` returns the expected `functions` / `classes` / `imports` / `exports`, and `extractCallGraph` returns `{ caller: "main", callee: "increment" }` plus `{ caller: "main", callee: "Counter" }`.

## Versioning

- [x] Version bumped 2.7.7 → 2.8.0 in all five manifests (per CLAUDE.md): `understand-anything-plugin/package.json`, `understand-anything-plugin/.claude-plugin/plugin.json`, `.claude-plugin/plugin.json`, `.cursor-plugin/plugin.json`, `.copilot-plugin/plugin.json`.


thejesh23:
Opened #436 to retroactively satisfy the README's *"please open an issue first for major changes"* guideline — apologies for the order. The issue summarizes the three open Dart PRs (#348, #415, this one) and asks maintainers to pick a shipping approach before reviewing diffs. Happy to close this PR in favour of either alternative if that's the call.

thejesh23:
@Lum1104 thanks for the direction. Addressed in two new commits:

- **`e4301ae`** — trimmed the process docs from `docs/superpowers/` (the design + plan files were personal brainstorm artifacts, not repo documentation). `BUILD.md` next to the vendored wasm stays as that's actual repo-level provenance.
- **`68777fe`** — broadened extractor coverage per your asks, pulling the stronger pieces from #348 and #415 where they fit cleanly:
  - **Optional + named parameters**: `extractParams` walks `optional_formal_parameters` (the Dart grammar uses one wrapper for both `[...]` and `{...}` — the leading token differs but the contained `formal_parameter` children are the same).
  - **`this.field` / `super.field` initializers**: new `extractParamName` helper unwraps `constructor_param` / `super_formal_parameter` and surfaces just the field name.
  - **Getters + setters**: `collectClassBody` now routes both `getter_signature` and `setter_signature` in both the concrete (`method_signature > …`) and abstract (`declaration > …`) shapes. The previous limitation assertion (`methods).not.toContain("value")`) flipped to positive `.toContain("value")`.
  - **Import edge cases**: added tests for `dart:` SDK URIs, multi-import declaration-order preservation, and `export 'foo.dart' show X`.
  - **Comma-list fields**: `int a, b, c;` now has explicit test coverage (the existing walk already handled it; locking it in).
  - Underscore-prefix visibility carries through to getters/setters via the existing `isExported` gate in `pushMethod` — explicit test added.

**Verification (all green)**

| | Before | After |
|---|---|---|
| Dart-extractor tests | 28 | **41** (+13) |
| Core suite | 720 | **733** |
| `pnpm lint` | ✅ | ✅ |
| `pnpm --filter @understand-anything/core build` | ✅ | ✅ |
| `pnpm --filter @understand-anything/skill build` | ✅ | ✅ |

Branch is up-to-date with `main` (0 commits behind). Call graph stays in scope as you wanted. Ready for another look when you have a moment.

Lum1104:
Hi, can you update the lock file too

Lum1104:
@codex review this

thejesh23:
@Lum1104 root lockfile + both codex findings addressed.

**`f2d6b99`** — synced the root `pnpm-lock.yaml`. The previous commits ran `pnpm install` inside `understand-anything-plugin/` so the workspace dep only landed in the inner lockfile; CI uses the root one with frozen-lockfile so the install would have failed. Codex's P1 finding — fixed.

**`a555f4d`** — addressed both codex P2 findings:

| Codex flag | What was wrong | Fix |
|---|---|---|
| `const Foo()` / `new Foo()` dropped from call graph | Grammar emits `const_object_expression` / `new_expression` with `arguments` directly — bypasses the `selector > argument_part` shape | Added a dedicated branch in `walkNode` that emits the inner `type_identifier` as the callee. Critical for Flutter `runApp(const MyApp())` patterns. |
| Calls inside ctor/factory/getter/setter bodies dropped | `method_signature` wraps `getter_signature` / `setter_signature` / `constructor_signature` / `factory_constructor_signature` when the member has a body — not `function_signature`. `pendingName` stayed null, sibling `function_body` walked with empty stack. | Dispatch across all five signature shapes; use `constructorName` for the (factory) ctor pair so the qualified name matches what `collectClassBody` pushes (`Foo.fromString`, etc.) |

**6 new tests** locking the behavior in: `const MyApp()` in a Flutter-shaped fixture, `new Counter()`, calls inside getter body, calls inside setter body, calls inside constructor body, calls inside factory body attributed to `Class.named`.

**Verification (all green)**

| | Before | After |
|---|---|---|
| Dart-extractor tests | 41 | **47** (+6) |
| Core suite | 733 | **739** |
| Root lockfile in sync | ❌ | ✅ |
| `pnpm lint` | ✅ | ✅ |
| `pnpm --filter @understand-anything/core build` | ✅ | ✅ |
| `pnpm --filter @understand-anything/skill build` | ✅ | ✅ |

cc @chatgpt-codex-connector — thanks for catching both P2s, they were real gaps.

PR #442, opened 2026-06-14
fix(typescript-extractor): capture parenless arrow params and abstract classes

tirth8205:
## Problem

Two structural-analysis gaps in the TypeScript/JavaScript extractor:

1. **Parenless single-param arrow functions** — `const g = x => doThing(x)` puts the lone parameter under the `arrow_function`'s `parameter` (singular) field, not inside a `formal_parameters` node. `extractVariableDeclarations` only looked for `parameters`/`formal_parameters`, so the parameter was silently dropped (`params: []`). This form is ubiquitous in map/filter callbacks.

2. **Abstract classes** — `abstract class Foo {}` parses as `abstract_class_declaration`, not `class_declaration`. Neither `processTopLevelNode` nor `processExportStatement` matched that node type, so abstract classes disappeared entirely: no class entry, no methods/properties, and `export abstract class X {}` produced no export entry either. Whole base-class/service hierarchies vanished from the graph.

## Fix

- In `extractVariableDeclarations`, read `valueNode.childForFieldName("parameter")` first and use it as the single param when present; otherwise fall back to the existing `parameters`/`formal_parameters` lookup.
- Add `case "abstract_class_declaration"` alongside `case "class_declaration"` in both `processTopLevelNode` and `processExportStatement`. `extractClass` already locates the name via `type_identifier`/`identifier` and reads `class_body`, so it works unchanged for the abstract variant.

## Testing

- Added `packages/core/src/plugins/extractors/__tests__/typescript-extractor.test.ts` (modeled on the sibling extractor tests, loading the `tree-sitter-typescript` WASM grammar) with cases for the parenless arrow param and for exported/non-exported abstract classes.
- The new assertions **fail before** the fix (params `[]`; classes/exports empty) and **pass after**.
- `vitest run` for the core package is green (35 files, 697 tests, no regressions); `tsc --noEmit` exits 0; ESLint on the changed files is clean.

### DIFF
--- a/understand-anything-plugin/packages/core/src/plugins/extractors/__tests__/typescript-extractor.test.ts
+++ b/understand-anything-plugin/packages/core/src/plugins/extractors/__tests__/typescript-extractor.test.ts
@@ -1,0 +1,99 @@
+import { describe, it, expect, beforeAll } from "vitest";
+import { createRequire } from "node:module";
+import { TypeScriptExtractor } from "../typescript-extractor.js";
+
+const require = createRequire(import.meta.url);
+
+// Load tree-sitter + TypeScript grammar once
+let Parser: any;
+let Language: any;
+let tsLang: any;
+
+beforeAll(async () => {
+  const mod = await import("web-tree-sitter");
+  Parser = mod.Parser;
+  Language = mod.Language;
+  await Parser.init();
+  const wasmPath = require.resolve(
+    "tree-sitter-typescript/tree-sitter-typescript.wasm",
+  );
+  tsLang = await Language.load(wasmPath);
+});
+
+function parse(code: string) {
+  const parser = new Parser();
+  parser.setLanguage(tsLang);
+  const tree = parser.parse(code);
+  const root = tree.rootNode;
+  return { tree, parser, root };
+}
+
+describe("TypeScriptExtractor", () => {
+  const extractor = new TypeScriptExtractor();
+
+  it("has correct languageIds", () => {
+    expect(extractor.languageIds).toEqual(["typescript", "javascript"]);
+  });
+
+  describe("extractStructure - arrow functions", () => {
+    it("captures the param of a parenless single-param arrow function", () => {
+      const { tree, parser, root } = parse(`const g = x => doThing(x);`);
+      const result = extractor.extractStructure(root);
+
+      expect(result.functions).toHaveLength(1);
+      expect(result.functions[0].name).toBe("g");
+      // Today: params is [] because the lone param lives under field
+      // `parameter` (singular) and is not wrapped in `formal_parameters`.
+      expect(result.functions[0].params).toEqual(["x"]);
+
+      tree.delete();
+      parser.delete();
+    });
+
+    it("still captures parenthesised arrow params", () => {
+      const { tree, parser, root } = parse(
+        `const add = (a, b) => a + b;`,
+      );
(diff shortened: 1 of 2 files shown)


thejesh23 (commented):
**1. Abstract method signatures still dropped from `methods`**
`extractClass` only matches `method_definition`, but abstract members parse as `abstract_method_signature` (e.g. `abstract run(): void;`). For the common case of an abstract base class declaring its contract, the resulting `classes[].methods` is empty or partial — exactly the hierarchy this PR is trying to recover. The new abstract-class test uses a concrete `run() {}` body, so this gap is not exercised.

**2. Decorated abstract classes**
TS decorators (`@Injectable() abstract class X {}`) wrap the class in a `decorator` + `class_declaration`/`abstract_class_declaration` pair, and when combined with `export` the node layout shifts again. Neither `processTopLevelNode` nor `processExportStatement` walks past a leading `decorator` sibling, so decorated abstract services (very common in Nest/Angular code) still disappear. Worth at least one test covering `@injectable() abstract class X {}`.

**3. Sibling Dart PR has the same shape**
PR #435 adds the Dart extractor and Dart also has `abstract class` plus parenless single-param closures via `=>`; the same two omissions (abstract-method-signature extraction, decorator/annotation wrapping) are likely to recur there. Worth aligning the fixes/tests across the two extractors before both land.

Nit: `singleParam.text` will inline any incidental whitespace/comments between the bare identifier and `=>`; harmless today but `singleParam.childForFieldName("name")?.text ?? singleParam.text` would be more defensive.

tirth8205:
**1. Abstract method signatures dropped from `methods`** — Fixed. `extractClass` now accepts `abstract_method_signature` and `method_signature` in addition to `method_definition`. Worth noting the node type depends on the `abstract` keyword: `abstract run(): void;` parses as `abstract_method_signature`, but a bare `run(): void;` body member (also valid in an abstract class) parses as `method_signature` — handling only the former would still drop the latter, so both are accepted. Added tests for a mixed abstract/concrete class (`methods === ['run','helper']`) and a purely-abstract contract (`['find','save']`).

**2. Decorated abstract classes** — Verified against tree-sitter-typescript 0.23.2: for `@Injectable() abstract class X {}` the top-level node is `abstract_class_declaration` with the decorator as an *inner child*, not a leading sibling; for `@Injectable() export abstract class X {}` and `export @Injectable() abstract class X {}` the `export_statement` carries `abstract_class_declaration` as a child. All three are already dispatched correctly (the name lookup `.find(type_identifier|identifier)` skips the decorator), so no code change was needed. Added regression tests for the decorated and decorated+exported forms to lock this.

**3. Dart parity (#435)** — Left as a separate follow-up. #435 is already merged, so the two can't land together; Dart is a distinct extractor with a different grammar and would need its own node-shape verification rather than assuming parity. Out of scope for this TS-only PR.

**Nit (`singleParam.text`)** — Keeping as-is. A node's `.text` is its own source span, so for `const g = x /* c */ => x` the `parameter` field points at the `identifier` `x` and `.text` is exactly `'x'` — no incidental whitespace/comments are inlined. The suggested `singleParam.childForFieldName('name')?.text ?? singleParam.text` is a no-op here: `singleParam` *is* the bare identifier and has no `name` field, so the first branch is always `undefined` and it falls through to `.text` regardless.

Full core suite green (701/701), tsc clean.

PR #446, opened 2026-06-14
fix(kotlin-extractor): collect members declared in enum class bodies

tirth8205:
## Problem

An `enum class` body is parsed by tree-sitter as an `enum_class_body` node, **not** `class_body`. `extractClassDeclaration` in `kotlin-extractor.ts` only looked up `findChild(declNode, "class_body")`, which returns `null` for an enum class, so `collectClassBody` was never invoked.

As a result, every method and property declared inside an enum class body was silently dropped:
- not added to the class's `methods[]`
- not added to the top-level `functions[]`
- not added to `exports[]`

For example, `enum class Direction { NORTH, SOUTH; fun opposite(): Direction = NORTH }` yielded `classes: [{ name: "Direction", methods: [], properties: [] }]` and `functions: []` — the `opposite` method vanished entirely. (Enum primary-constructor `val`/`var` properties still worked, since those come from `primary_constructor`, not the body.)

## Fix

Look up either body variant before collecting members:

```ts
const body =
  findChild(declNode, "class_body") ??
  findChild(declNode, "enum_class_body");
```

`collectClassBody` already only reacts to `function_declaration` / `property_declaration` / `object_declaration` children, so the `enum_entry` children of an enum body are harmlessly ignored — no other change is needed.

## Testing

- Added a test `collects methods declared inside an enum class body` to `kotlin-extractor.test.ts`. It **fails before** the fix (`expected [] to include 'opposite'`) and **passes after**.
- The full core test suite is green: 693 passed.
- Typecheck (`tsc --noEmit`) exits 0 and ESLint on both changed files exits 0.

### DIFF
--- a/understand-anything-plugin/packages/core/src/plugins/extractors/__tests__/kotlin-extractor.test.ts
+++ b/understand-anything-plugin/packages/core/src/plugins/extractors/__tests__/kotlin-extractor.test.ts
@@ -173,6 +173,85 @@ fun three(): String = ""
       expect(result.classes[0].name).toBe("Direction");
       expect(result.classes[0].methods).toContain("opposite");
       expect(result.functions.map((f) => f.name)).toContain("opposite");
+      // A default-public body member is exported (Kotlin default visibility).
+      expect(result.exports.map((e) => e.name)).toContain("opposite");
+
+      tree.delete();
+      parser.delete();
+    });
+
+    it("surfaces enum constants (enum_entry) as properties", () => {
+      const { tree, parser, root } = parse(`enum class Direction {
+    NORTH, SOUTH;
+    fun opposite(): Direction = NORTH
+}
+`);
+      const result = extractor.extractStructure(root);
+
+      // The enum constants are the canonical members of the enum.
+      expect(result.classes[0].properties).toEqual(
+        expect.arrayContaining(["NORTH", "SOUTH"]),
+      );
+      // Enum constants are public and therefore exported.
+      expect(result.exports.map((e) => e.name)).toEqual(
+        expect.arrayContaining(["NORTH", "SOUTH"]),
+      );
+
+      tree.delete();
+      parser.delete();
+    });
+
+    it("collects methods from a companion object inside a class body", () => {
+      const { tree, parser, root } = parse(`class Widget {
+    companion object {
+        fun create(): Widget = Widget()
+    }
+}
+`);
+      const result = extractor.extractStructure(root);
+
+      // The companion's function is surfaced on the enclosing class and at
+      // the top level (mirrors the method-into-functions[] convention).
+      expect(result.classes[0].methods).toContain("create");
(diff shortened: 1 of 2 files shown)


thejesh23 (commented):
**1. Enum constants themselves are not surfaced as members.**
`collectClassBody` only reacts to `function_declaration` / `property_declaration` / `object_declaration`, so `enum_entry` children (NORTH, SOUTH) are silently dropped. For a fix targeted at enum bodies, omitting the entries themselves is the larger gap — they're the canonical "members" of an enum and would be the natural thing to land in `properties[]` (parallels the val/var primary-constructor handling). My in-flight Dart PR #435 will hit the same modeling question for `enum_constant_declaration`.

**2. Companion object inside an enum body is not exercised.**
`object_declaration` is handled, but `companion_object` is a distinct node type in tree-sitter-kotlin and won't match. An enum class with `companion object { fun create(): Direction = NORTH }` likely still loses `create`. Worth a test, and a separate `findChild(body, "companion_object")` branch if confirmed.

**3. Test gap: properties and export are not asserted.**
The new test only checks `classes[0].methods` and `functions`. It does not assert `result.exports` contains `opposite` (the default-public path through `isExported` is the actual regression surface), nor that the enum entries appear anywhere. Consider adding both, plus a `private fun` case to lock down visibility.

Nit: the comment "enum class body is parsed as `enum_class_body`" is the load-bearing fact — worth linking the tree-sitter-kotlin grammar node name in the comment so the next maintainer doesn't have to re-derive it.

tirth8205:
All four points addressed in 78e5da4. Node shapes were confirmed against the real tree-sitter-kotlin parse tree before implementing.

**1. Enum constants now surfaced.** `enum_entry` is a direct named child of `enum_class_body`, each carrying its name as an `identifier` child. `collectClassBody` now pushes each entry into `properties[]` and `exports[]` (enum constants are always public). Test: `surfaces enum constants (enum_entry) as properties` asserts both `NORTH`/`SOUTH` in `properties` and in `exports`.

**2. Companion object handled.** Confirmed `companion_object` is its own node type — a sibling of `class_member_declaration`, not a subtype of `object_declaration` — so it never matched the existing branch. Added an `else if (member.type === "companion_object")` branch that surfaces the (optional) companion name as a property and recurses into its `class_body`, so companion functions like `create` land in `methods[]`/`functions[]`/`exports[]`. Tests cover a companion inside both a plain class body and an enum class body.

**3. Export/visibility gap closed.** The enum test now asserts `exports` contains the default-public body member `opposite`. Added a `does NOT export a private function declared inside a class body` test asserting a body-level `private fun` is excluded from `exports` while still appearing in the class's `methods[]`.

**Nit.** The `enum_class_body` comment now names the grammar source (`@tree-sitter-grammars/tree-sitter-kotlin`, `enum_class_body` rule).

Full core suite green (697/697), `tsc --noEmit` clean.

PR #448, opened 2026-06-14
fix(tests): make scan-project fixtures hermetic against host global gitignore (#427)

### LINKED ISSUE #427: bug: scan-project tests fail when the contributor's global gitignore excludes .env (fixture scans are not hermetic)
### What happened?

Ran the skill tests (`pnpm test`, or file-scoped: `./node_modules/.bin/vitest run tests/skill/understand/test_scan_project.test.mjs`) on a development machine whose **global gitignore contains `.env` and `.env.local`**. One test fails, and it fails with a TypeError rather than an assertion:

```
FAIL  tests/skill/understand/test_scan_project.test.mjs > scan-project.mjs — category assignment (project-scanner.md Step 4) > dotfile configs (.env, .env.local, .env.production) map to config + env language
TypeError: Cannot read properties of undefined (reading 'fileCategory')
 ❯ tests/skill/understand/test_scan_project.test.mjs:401:33
```

Expected: the suite passes regardless of the contributor's personal git configuration.

CI never sees this because `ubuntu-latest` runners ship with no global gitignore. It only bites contributor machines, and `.env` is one of the most common global-ignore entries in the wild.

### Root cause

`scan-project.mjs` prefers git enumeration: `enumerateViaGit()` spawns `git ls-files -z -co --exclude-standard`. Per gitignore(5), `--exclude-standard` honors three ignore sources: the repo's `.gitignore`, `.git/info/exclude`, and the host's `core.excludesFile`. The test helper `runScript()` spawns the scanner with a fully inherited environment, so the host's global ignore applies inside every fixture repo. With `.env` globally ignored, the fixture's `.env`/`.env.local` never reach the scanner output, `byPath(r.output, '.env')` returns `undefined`, and the test dies on `undefined.fileCategory`.

The walker fallback never consults gitignore, which makes this look spooky during diagnosis: the identical tree scans fine with `gitInit: false`.

### Minimal reproduction

No need to touch your real config; simulate the global ignore via `XDG_CONFIG_HOME`:

```bash
export XDG_CONFIG_HOME="$(mktemp -d)"
mkdir -p "$XDG_CONFIG_HOME/git"
printf '.env\n.env.local\n' > "$XDG_CONFIG_HOME/git/ignore"

repo="$(mktemp -d)"; git -C "$repo" init -q
touch "$repo/.env" "$repo/.env.local" "$repo/.env.production"

git -C "$repo" ls-files -co --exclude-standard
# prints only: .env.production
```

Run the vitest file under those same two exports and the dotfile-configs test fails exactly as above.

### The trap: nulling the config files is NOT enough

The obvious hermeticity fix, `GIT_CONFIG_GLOBAL=/dev/null GIT_CONFIG_SYSTEM=/dev/null`, does **not** work:

```bash
$ GIT_CONFIG_GLOBAL=/dev/null GIT_CONFIG_SYSTEM=/dev/null git -C "$repo" ls-files -co --exclude-standard
.env.production                      # still broken

$ GIT_CONFIG_GLOBAL=/dev/null git config --get core.excludesFile; echo "exit=$?"
exit=1                               # the config setting is gone...

$ GIT_CONFIG_GLOBAL=/dev/null GIT_CONFIG_SYSTEM=/dev/null git -C "$repo" check-ignore -v .env
~/.config/git/ignore:7:.env	.env    # ...yet the ignore file still applies
```

That is because `core.excludesFile` has a built-in default independent of any config file. git-config(1):

> Defaults to `$XDG_CONFIG_HOME/git/ignore`. If `$XDG_CONFIG_HOME` is either not set or empty, `$HOME/.config/git/ignore` is used instead.

So the knob itself must be overridden. Git's environment-config mechanism does that without touching the scanner's command line, keeping the fix entirely inside the tests.

### Suggested fix (verified)

```diff
diff --git a/tests/skill/understand/test_scan_project.test.mjs b/tests/skill/understand/test_scan_project.test.mjs
index 65d96c8..fe56a5c 100644
--- a/tests/skill/understand/test_scan_project.test.mjs
+++ b/tests/skill/understand/test_scan_project.test.mjs
@@ -19,6 +19,26 @@ const SCRIPT = resolve(
   '../../../understand-anything-plugin/skills/understand/scan-project.mjs',
 );
 
+/**
+ * Hermetic git environment for fixture work: host-level git config must not
+ * leak into tests. Developers commonly have `.env` in their global gitignore;
+ * scan-project.mjs enumerates via `git ls-files -co --exclude-standard`,
+ * which honors core.excludesFile, so fixture files silently vanish from the
+ * scan and byPath() lookups explode. Nulling the config files alone is NOT
+ * enough: core.excludesFile *defaults* to `$XDG_CONFIG_HOME/git/ignore`
+ * (falling back to `~/.config/git/ignore`) even when no config sets it, so
+ * the knob itself is overridden via git's environment-config mechanism
+ * (GIT_CONFIG_COUNT: git >= 2.31; GIT_CONFIG_GLOBAL/SYSTEM: git >= 2.32).
+ */
+const HERMETIC_GIT_ENV = {
+  ...process.env,
+  GIT_CONFIG_GLOBAL: '/dev/null',
+  GIT_CONFIG_SYSTEM: '/dev/null',
+  GIT_CONFIG_COUNT: '1',
+  GIT_CONFIG_KEY_0: 'core.excludesFile',
+  GIT_CONFIG_VALUE_0: '/dev/null',
+};
+
 /**
  * Build a project tree from a `{ relPath: contents }` object. Creates parent
  * directories as needed. Initializes a real git repo so the script's preferred
@@ -36,7 +56,11 @@ function setupTree(files, { gitInit = true } = {}) {
     // `git ls-files -co --exclude-standard` returns BOTH cached and others
     // (modulo gitignore), so an `add` is unnecessary for our tests — the
     // bare repo init is enough for ls-files to enumerate.
-    const init = spawnSync('git', ['init', '-q'], { cwd: root, encoding: 'utf-8' });
+    const init = spawnSync('git', ['init', '-q'], {
+      cwd: root,
+      encoding: 'utf-8',
+      env: HERMETIC_GIT_ENV,
+    });
     if (init.status !== 0) {
       // CI without git: continue without it; the walker fallback will fire.
     }
@@ -66,6 +90,7 @@ function runScript(projectRoot) {
   const outputPath = join(outputDir, 'scan-output.json');
   const result = spawnSync('node', [SCRIPT, projectRoot, outputPath], {
     encoding: 'utf-8',
+    env: HERMETIC_GIT_ENV,
   });
   let output = null;
   try {
```

Verification on the affected machine: without the patch the file runs 31 passed / 1 failed (the TypeError above); with the patch, 32/32 pass (vitest exit code 0). The fixture `git init` gets the same env for completeness: a host `init.templateDir` could otherwise plant `.git/info/exclude` entries, which is the same class of leak.

Version floors, from the git release notes: the `GIT_CONFIG_COUNT`/`GIT_CONFIG_KEY_n`/`GIT_CONFIG_VALUE_n` mechanism landed in git 2.31.0 and `GIT_CONFIG_GLOBAL`/`GIT_CONFIG_SYSTEM` in 2.32.0 (June 2021), comfortably below anything that can run this repo's Node >= 22 toolchain.

Alternative considered: passing `-c core.excludesFile=/dev/null` inside `enumerateViaGit()` also fixes the test (verified), but that changes production behavior to solve a test-hermeticity problem, so the env route in the test helper seems strictly better. Scope note: of the three skill scripts under test, only `scan-project.mjs` shells out to git, so this one helper is the entire blast radius.

### Two small adjacent observations

- `byPath()`'s docstring says callers should `expect(byPath(...)).toBeDefined()` first; the dotfile test indexes the result directly, which is why this surfaced as a TypeError instead of a readable assertion failure. Adding the guard there would make the next config-leak diagnosis faster.
- Deliberately out of scope here: whether production scans *should* let a user's personal global gitignore hide files like `.env` from analysis (the walker fallback and git path currently disagree about that). This issue is only about making the tests independent of contributor machine config.

### Environment

- Plugin version: 2.7.6, repo at `fb1356a`
- Platform: Claude Code (CLI), plugin marketplace checkout
- OS + Node: macOS 26.2 (arm64), Node v24.15.0, vitest 3.2.4, git 2.50.1 (Apple Git-155)

tirth8205:
## Problem

`scan-project.mjs` enumerates files via `git ls-files -co --exclude-standard`, which honors the host's `core.excludesFile` (the global gitignore). Many contributors keep `.env` / `.env.local` in their global gitignore. When they do, the test fixtures' dotfiles never reach the scanner, `byPath(r.output, '.env')` returns `undefined`, and the "dotfile configs" test dies with:

```
TypeError: Cannot read properties of undefined (reading 'fileCategory')
```

CI (ubuntu-latest) has no global gitignore, so it never reproduces this — it only bites local contributors. Nulling `GIT_CONFIG_GLOBAL` / `GIT_CONFIG_SYSTEM` is **not** enough, because `core.excludesFile` defaults to `$XDG_CONFIG_HOME/git/ignore` (or `~/.config/git/ignore`) even with no config file present — the knob itself must be overridden.

## Fix

Add a `HERMETIC_GIT_ENV` constant that nulls the global/system git config and overrides `core.excludesFile` via the `GIT_CONFIG_COUNT`/`GIT_CONFIG_KEY_0`/`GIT_CONFIG_VALUE_0` mechanism, then pass it as `env` to **both** spawnSync calls — the `git init` in `setupTree()` and the `node scan-project.mjs` invocation in `runScript()`. This makes the fixtures hermetic against any host git configuration. The dotfile test's lookup is also guarded with `toBeDefined()` for a clearer failure message if this ever regresses.

Only the test file `tests/skill/understand/test_scan_project.test.mjs` changes; the scanner itself is untouched.

## Testing

Reproduced the bug by simulating a host global gitignore that ignores `.env`:

```sh
export XDG_CONFIG_HOME="$(mktemp -d)"; mkdir -p "$XDG_CONFIG_HOME/git"
printf '.env\n.env.local\n' > "$XDG_CONFIG_HOME/git/ignore"
```

- **Before the fix**, with that env: the suite fails 1/32 with the `TypeError ... reading 'fileCategory'` at the `.env` dotfile test.
- **After the fix**, with that env: 32/32 pass.
- **After the fix**, without that env (clean): 32/32 pass.

Lint (`eslint`) and core typecheck (`tsc --noEmit`) are green; the package test suite (`vitest run tests/skill/understand/test_scan_project.test.mjs`) passes 32/32 both with and without the simulated global ignore.

## Hardening the regression (follow-up commit)

The original dotfile-config fixture was dotfile-only. When a host global gitignore hides every dotfile, `git ls-files` returns an empty list, so `scan-project.mjs` falls back to its (gitignore-unaware) recursive walker and re-discovers the dotfiles — masking the #427 leak even without the hermetic env. Adding a non-dotfile sibling (`keep.ts`) to that fixture keeps `git ls-files` output non-empty, forcing the git enumeration path so the regression reliably **fails without** the hermetic env (`expected undefined to be defined`) and **passes with** it. Verified red→green:

- Unpatched env + simulated `XDG_CONFIG_HOME` global ignore → the dotfile test FAILS.
- Unpatched env + no simulated ignore → 32/32 pass (explains why CI never caught it).
- Patched (`HERMETIC_GIT_ENV`) → 32/32 pass both with and without the simulated ignore.

Full root vitest suite (16 files, 207 tests) and `eslint` on the changed file are green.

Fixes #427

### DIFF
--- a/tests/skill/understand/test_scan_project.test.mjs
+++ b/tests/skill/understand/test_scan_project.test.mjs
@@ -20,6 +20,26 @@ const SCRIPT = resolve(
 );
 
 /**
+ * scan-project.mjs enumerates files via `git ls-files -co --exclude-standard`,
+ * which honors the host's global gitignore (core.excludesFile). Many
+ * contributors keep `.env`/`.env.local` in their global gitignore, which would
+ * then hide the fixture dotfiles from the scanner and break tests that assert
+ * on them (e.g. the `.env` dotfile-config test) — see #427. Nulling
+ * GIT_CONFIG_GLOBAL/SYSTEM is NOT enough: core.excludesFile defaults to
+ * $XDG_CONFIG_HOME/git/ignore (or ~/.config/git/ignore) even with no config
+ * file, so the knob itself must be overridden. Passing this env to every git
+ * invocation makes the fixtures hermetic against any host config.
+ */
+const HERMETIC_GIT_ENV = {
+  ...process.env,
+  GIT_CONFIG_GLOBAL: '/dev/null',
+  GIT_CONFIG_SYSTEM: '/dev/null',
+  GIT_CONFIG_COUNT: '1',
+  GIT_CONFIG_KEY_0: 'core.excludesFile',
+  GIT_CONFIG_VALUE_0: '/dev/null',
+};
+
+/**
  * Build a project tree from a `{ relPath: contents }` object. Creates parent
  * directories as needed. Initializes a real git repo so the script's preferred
  * `git ls-files` enumeration path is exercised — tests that need the walker
@@ -36,7 +56,11 @@ function setupTree(files, { gitInit = true } = {}) {
     // `git ls-files -co --exclude-standard` returns BOTH cached and others
     // (modulo gitignore), so an `add` is unnecessary for our tests — the
     // bare repo init is enough for ls-files to enumerate.
-    const init = spawnSync('git', ['init', '-q'], { cwd: root, encoding: 'utf-8' });
+    const init = spawnSync('git', ['init', '-q'], {
+      cwd: root,
+      encoding: 'utf-8',
+      env: HERMETIC_GIT_ENV,
+    });
     if (init.status !== 0) {
       // CI without git: continue without it; the walker fallback will fire.
     }
@@ -66,6 +90,7 @@ function runScript(projectRoot) {
   const outputPath = join(outputDir, 'scan-output.json');
   const result = spawnSync('node', [SCRIPT, projectRoot, outputPath], {
     encoding: 'utf-8',
+    env: HERMETIC_GIT_ENV,
   });
   let output = null;
   try {
@@ -398,10 +423,15 @@ describe('scan-project.mjs — category assignment (project-scanner.md Step 4)',
     const r = runScript(projectRoot);
     expect(r.status).toBe(0);
     for (const p of ['.env', '.env.local', '.env.production']) {
-      expect(byPath(r.output, p).fileCategory).toBe('config');
+      const entry = byPath(r.output, p);
+      // Guard for a clearer failure than "Cannot read properties of
+      // undefined" if a host global gitignore ever hides these dotfiles
+      // (see HERMETIC_GIT_ENV above / #427).
+      expect(entry).toBeDefined();
+      expect(entry.fileCategory).toBe('config');
       // LANGUAGE_BY_EXT['.env'] -> 'config' (the language id itself; not
       // a typo — the language for env files is the 'config' bucket).
-      expect(byPath(r.output, p).language).toBe('config');
+      expect(entry.language).toBe('config');
     }
   });
 });
--- a/tests/skill/understand/test_scan_project.test.mjs
+++ b/tests/skill/understand/test_scan_project.test.mjs
@@ -416,6 +416,13 @@ describe('scan-project.mjs — category assignment (project-scanner.md Step 4)',
   // review on PR #204.
   it('dotfile configs (.env, .env.local, .env.production) map to config + env language', () => {
     projectRoot = setupTree({
+      // A non-dotfile sibling keeps `git ls-files` output non-empty so the
+      // git enumeration path is actually exercised. Without it, a host global
+      // gitignore that hides every dotfile would make `git ls-files` return
+      // nothing, silently tripping the walker fallback (which is unaware of
+      // gitignore) and masking the #427 leak. With a sibling present, the git
+      // path is taken and the dotfiles must survive on their own merits.
+      'keep.ts': 'export const x = 1;\n',
       '.env': 'API_KEY=abc\n',
       '.env.local': 'LOCAL=1\n',
       '.env.production': 'PROD=1\n',
--- a/tests/skill/understand/test_scan_project.test.mjs
+++ b/tests/skill/understand/test_scan_project.test.mjs
@@ -8,7 +8,7 @@ import {
   chmodSync,
   existsSync,
 } from 'node:fs';
-import { tmpdir } from 'node:os';
+import { tmpdir, devNull } from 'node:os';
 import { join, dirname, resolve } from 'node:path';
 import { spawnSync } from 'node:child_process';
 import { fileURLToPath } from 'node:url';
@@ -29,14 +29,17 @@ const SCRIPT = resolve(
  * $XDG_CONFIG_HOME/git/ignore (or ~/.config/git/ignore) even with no config
  * file, so the knob itself must be overridden. Passing this env to every git
  * invocation makes the fixtures hermetic against any host config.
+ *
+ * `os.devNull` is used instead of a literal '/dev/null' so the null sink is
+ * valid on Windows too (`NUL`), where contributors also run these tests.
  */
 const HERMETIC_GIT_ENV = {
   ...process.env,
-  GIT_CONFIG_GLOBAL: '/dev/null',
-  GIT_CONFIG_SYSTEM: '/dev/null',
+  GIT_CONFIG_GLOBAL: devNull,
+  GIT_CONFIG_SYSTEM: devNull,
   GIT_CONFIG_COUNT: '1',
   GIT_CONFIG_KEY_0: 'core.excludesFile',
-  GIT_CONFIG_VALUE_0: '/dev/null',
+  GIT_CONFIG_VALUE_0: devNull,
 };
 
 /**


thejesh23 (commented):
**1. `/dev/null` paths are not portable to Windows contributors**

The PR's stated purpose is unbreaking *local* contributor runs, but `/dev/null` doesn't exist on Windows — git on Windows will fail to open it as the `core.excludesFile` (and as the global/system config). On Windows the hermetic env will itself error out `git ls-files`, silently triggering the walker fallback (which is what the fix tries to avoid). Consider `os.devNull` (`NUL` on win32) or an empty tmp file created once per process.

**2. `GIT_CONFIG_NOSYSTEM` not set; system include paths still resolvable**

`GIT_CONFIG_SYSTEM=/dev/null` only redirects the *primary* system file; it does not stop git from reading `$(prefix)/etc/gitconfig` includes or honoring `core.attributesFile` / `core.hooksPath` from a system install. The canonical fully-hermetic recipe is `GIT_CONFIG_NOSYSTEM=1` plus `GIT_CONFIG_GLOBAL=/dev/null` plus the `GIT_CONFIG_COUNT` override. Worth adding for completeness, since #427's root cause is exactly this class of leakage.

**3. Only one fixture hardened against the silent-walker-fallback masking**

The follow-up addition of `keep.ts` to force the git enumeration path is applied solely to the dotfile-configs test. Other fixtures that are dotfile-only or sparse (e.g., the `.env` / config-related setups elsewhere in this file) still risk emptying `git ls-files` and falling through to the gitignore-unaware walker, where the same class of host-config bug would be masked again. Either generalize `setupTree` to inject a sentinel non-dotfile, or assert `r.stderr`/source-of-truth is git in tests that care.

tirth8205:
**1. `/dev/null` not portable to Windows** — Fixed. Imported `devNull` from `node:os` and replaced the three `/dev/null` literals in `HERMETIC_GIT_ENV` (`GIT_CONFIG_GLOBAL`, `GIT_CONFIG_SYSTEM`, `GIT_CONFIG_VALUE_0`) with it, so the null sink resolves to `NUL` on win32 and `/dev/null` on POSIX. Added a JSDoc note explaining the cross-platform rationale. The existing 32 tests in this file (and the full 102-test skill suite) stay green.

**2. `GIT_CONFIG_NOSYSTEM` not set** — Not changed. Env-injected `GIT_CONFIG_*` overrides take final precedence over any config file (system, its includes, and global), so the `GIT_CONFIG_COUNT`/`KEY_0`/`VALUE_0` override pins `core.excludesFile` to the null sink regardless of system includes — empirically `git config --get core.excludesFile` returns the override and `git ls-files -co --exclude-standard` still lists `.env` even when a system config tries to hide it. The other knobs cited (`core.attributesFile`, `core.hooksPath`) don't influence which untracked files `git ls-files -co --exclude-standard` enumerates (attributes affect diff/merge/eol; hooksPath affects hook lookup), so they can't reintroduce the dotfile-hiding bug. `GIT_CONFIG_NOSYSTEM=1` would be harmless defense-in-depth but is a no-op for this specific failure mode.

**3. Only one fixture hardened** — Already covered. `HERMETIC_GIT_ENV` is applied at the env level to both `setupTree`'s `git init` and `runScript`'s scanner subprocess, so no host global gitignore can hide any fixture dotfile in *any* test, not just the dotfile-config one. The `keep.ts` sentinel is deliberately scoped to that one test as a red→green regression anchor (so it reliably fails when the env is absent rather than silently passing via the walker fallback). Generalizing `setupTree` to inject a sentinel into every fixture would break the exact-count assertions (`totalFiles` of 30/31/150/151/501) and isn't needed given the env-level fix.

PR #455, opened 2026-06-14
fix(languages): add .mts/.cts extensions and *.spec.tsx test glob to the TypeScript config

tirth8205:
## Problem
- The TypeScript LanguageConfig declares `extensions: [".ts", ".tsx"]` but omits `.mts` and `.cts`, the first-class TypeScript module file extensions (TS 4.7+, used in ESM/nodenext projects — which this repo itself targets, see PR #405/#359). The JavaScript config correctly includes the analogous `.mjs` and `.cjs`. Because LanguageRegistry derives its extension map directly from `config.extensions`, and TreeSitterPlugin (tree-sitter-plugin.ts lines 66-72) builds `_extensionToLang` the same way, a `.mts` or `.cts` file resolves to NULL: it is not detected as TypeScript by `getByExtension`/`getForFile` and is skipped entirely by the tree-sitter extraction pipeline. Verified at runtime: `getByExtension('.mts')` -> NULL, `getByExtension('.cts')` -> NULL, while `getByExtension('.mjs')` -> 'javascript' and `getByExtension('.cjs')` -> 'javascript'. The tree-sitter-typescript grammar parses…
- The TypeScript config's test glob list is `tests: ["*.test.ts", "*.spec.ts", "*.test.tsx"]`. It covers `.test.tsx` (the React component test convention) but inconsistently omits `*.spec.tsx`, even though it covers both `.test`/`.spec` for plain `.ts`. The JavaScript config consistently lists both `*.test.js` and `*.spec.js`. The dashboard package is React + TypeScript and uses `.tsx` files, so `Component.spec.tsx` test files (a common naming style) are not recognized as tests by the file-pattern classifier, while the equivalent `Component.test.tsx` is. This causes spec-named TSX test files to be miscategorized as regular source.

## Fix
- Change line 6 to include the module extensions, mirroring the JavaScript config: extensions: [".ts", ".tsx", ".mts", ".cts"],
- Add the missing spec-tsx glob for parity with the test-tsx and spec-ts entries: tests: ["*.test.ts", "*.spec.ts", "*.test.tsx", "*.spec.tsx"],

## Testing

Found via a static correctness audit of the TypeScript language config.

### DIFF
--- a/understand-anything-plugin/packages/core/src/__tests__/language-registry.test.ts
+++ b/understand-anything-plugin/packages/core/src/__tests__/language-registry.test.ts
@@ -2,6 +2,7 @@ import { describe, it, expect } from "vitest";
 import { LanguageRegistry } from "../languages/language-registry.js";
 import { StrictLanguageConfigSchema } from "../languages/types.js";
 import { typescriptConfig } from "../languages/configs/typescript.js";
+import { javascriptConfig } from "../languages/configs/javascript.js";
 import { pythonConfig } from "../languages/configs/python.js";
 
 describe("LanguageRegistry", () => {
@@ -100,6 +101,13 @@ describe("LanguageRegistry", () => {
     it("recognizes both .test.tsx and .spec.tsx test files", () => {
       expect(typescriptConfig.filePatterns.tests).toContain("*.test.tsx");
       expect(typescriptConfig.filePatterns.tests).toContain("*.spec.tsx");
+    });
+  });
+
+  describe("javascript config test patterns", () => {
+    it("recognizes .jsx test files for parity with the .tsx patterns", () => {
+      expect(javascriptConfig.filePatterns.tests).toContain("*.test.jsx");
+      expect(javascriptConfig.filePatterns.tests).toContain("*.spec.jsx");
     });
   });
 
--- a/understand-anything-plugin/packages/core/src/__tests__/language-registry.test.ts
+++ b/understand-anything-plugin/packages/core/src/__tests__/language-registry.test.ts
@@ -72,6 +72,9 @@ describe("LanguageRegistry", () => {
       expect(registry.getByExtension(".h")?.id).toBe("c");
       expect(registry.getByExtension(".lua")?.id).toBe("lua");
       expect(registry.getByExtension(".js")?.id).toBe("javascript");
+      expect(registry.getByExtension(".mts")?.id).toBe("typescript");
+      expect(registry.getByExtension(".cts")?.id).toBe("typescript");
+      expect(registry.getForFile("src/server.mts")?.id).toBe("typescript");
     });
 
     it("has no duplicate extension mappings across configs", () => {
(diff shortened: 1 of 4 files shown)


thejesh23 (commented):

**1. `.d.ts` (and `.d.mts` / `.d.cts`) still skipped.** Declaration files are the most common TS-extension outside `.ts/.tsx`, and `getForFile("foo.d.ts")` matches on `.ts` so they already resolve — but the extractor then sees a pure-types file (no runtime exports, no call edges). Worth either explicitly listing `.d.ts` / `.d.mts` / `.d.cts` and gating them out of call-graph extraction, or adding a short comment in the config explaining that `.d.ts` is intentionally not separated. Otherwise the next person doing this audit lands a third PR.

**2. `.mts` / `.cts` now route to the plain TS grammar — confirm that's intended.** `tree-sitter-plugin.ts` lines 111-118 hard-code `ext === ".tsx" ? "tsx"` and falls through to `_extensionToLang.get(ext)` for everything else. So `.mts` and `.cts` will use `tree-sitter-typescript.wasm`, which is correct (they don't carry JSX). Worth a one-line code comment on that special-case so a future reader doesn't add `.mts` to the TSX branch by analogy with the `.mjs`-treated-as-JS pattern.

**3. Test glob shape only covers root-level patterns.** `*.test.tsx` / `*.spec.tsx` are bare globs (no `**/` prefix and no `__tests__/*.tsx`). Vitest's default `include` is `**/*.{test,spec}.?(c|m)[jt]s?(x)` and most repos colocate under `src/**/__tests__/`. Depending on how `filePatterns.tests` is consumed (couldn't find a hot caller in `src/` besides the new test) this may already not match `packages/foo/src/__tests__/bar.test.tsx`. Worth confirming the matcher is suffix-based, or extending to `**/__tests__/*.{ts,tsx}` for parity with how the layer-detector already treats `__tests__` as a test dir (`analyzer/layer-detector.ts:58`).

Nit: while you're here, JS config's `tests` is missing `*.test.jsx` / `*.spec.jsx` — same asymmetry you just fixed on the TS side, one config over.

tirth8205:
**1. `.d.ts` / `.d.mts` / `.d.cts`** — Added a comment in `typescript.ts` documenting that declaration files intentionally fall through to `.ts` / `.mts` / `.cts` and are parsed as ordinary types-only TypeScript (no separate call-graph gating). I did not add extractor gating: this behavior predates this PR (`.d.ts` already resolved via `.ts` before `.mts`/`.cts` were added), so special-casing the extractor is a behavior change outside the module-variant-extension scope here and risks regressing types-only parsing. Tracking the gating idea as a follow-up.

**2. `.mts` / `.cts` route to plain TS grammar** — Confirmed correct: `languageKeyFromPath` only special-cases `ext === ".tsx"` → synthetic `tsx` key; everything else (including `.mts`/`.cts`) falls through to `_extensionToLang.get(ext)` → `typescript` → `tree-sitter-typescript.wasm`. Added the requested one-line comment so a future reader doesn't add `.mts` to the `.tsx` branch by analogy with the `.mjs`-as-JS pattern.

**3. Test glob shape** — Not extending the globs. `LanguageConfig.filePatterns.tests` has no runtime matcher consumer anywhere in `src/` (verified by content grep — only config definitions and the registry test read it; it's `z.array(z.string())` declarative metadata). The `filePatterns` in `layer-detector.ts`/`llm-analyzer.ts` is an unrelated LLM-supplied layer-prefix field, and `__tests__` layer detection is handled independently by the hardcoded `LAYER_PATTERNS` in `layer-detector.ts`. Adding `**/__tests__/*.{ts,tsx}` would introduce unused, potentially misleading config rather than change any matching behavior.

**Nit (JS `.jsx` test patterns)** — Fixed for parity: added `*.test.jsx` / `*.spec.jsx` to `javascriptConfig.filePatterns.tests`, with a test locking it. (No functional effect today given #3, but it removes the cross-config asymmetry.)

Full core suite green (694 passed); `tsc --noEmit` clean.

PR #458, opened 2026-06-15
fix: harden supply-chain security — obfuscation scanner + advisory policy

SongotenU:
## Summary

Hardens supply-chain security following real malicious PRs observed in this repo:

- **PR #206**: obfuscated payload in `astro.config.mjs` via `global['!']` hijack + packed self-decoding blob
- **PR #198, #261**: same pattern from same author

### 1. SECURITY.md updates
- Direct link to [Security Advisories page](https://github.com/Egonex-AI/Understand-Anything/security/advisories/new)
- Email fallback (`security@egonex.ai`) if advisories unavailable
- Supply-chain attacks explicitly in-scope

### 2. Obfuscation scanner script (`.github/scripts/scan-obfuscation.sh`)
Detects patterns commonly found in malicious PRs:
- Packed/self-decoding JS blobs
- Large base64/hex-encoded strings  
- String-shuffle decoders
- `global['!']` / `global = global` hijacks
- Hidden eval/Function calls
- `require` rebinding via global
- Build-config file targeting (astro/vite/next configs)

### 3. Supply-chain CI workflow (`.github/workflows/supply-chain-scan.yml`)
- Runs on every PR (opened, synchronize, reopened)
- Generates full PR diff → runs obfuscation scanner
- **Non-blocking**: auto-comments when suspicious patterns found, but does NOT block merge (false-positive safety)
- Explicitly runs for fork PRs (highest risk vector)

## Test plan
- [x] Detects known attack pattern from PR #206 (tested with sample obfuscated payload)
- [x] CI workflow should trigger on next PR opened on upstream
- [x] Maintainer should enable Private Vulnerability Reporting in repo Settings → Code Security

### DIFF
--- a/.github/scripts/scan-obfuscation.sh
+++ b/.github/scripts/scan-obfuscation.sh
@@ -1,0 +1,134 @@
+#!/usr/bin/env bash
+# Supply-chain obfuscation scanner
+# Scans PR diffs for patterns commonly found in malicious supply-chain attacks:
+#   - Obfuscated code (packed, shuffled, base64-encoded, hex-encoded)
+#   - Suspicious global/prototype assignments in build config files
+#   - Hidden eval / Function constructor calls
+#   - Large diffs dominated by non-human-readable strings
+#
+# Exit 0 = clean. Exit 1 = suspicious patterns found (review required).
+
+set -euo pipefail
+
+RED='\033[0;31m'
+YELLOW='\033[1;33m'
+NC='\033[0m' # No Color
+
+WARNINGS=0
+
+check_pattern() {
+  local label="$1"
+  local pattern="$2"
+  local severity="${3:-WARN}"
+
+  if grep -n -P "$pattern" /tmp/pr-diff.txt 2>/dev/null; then
+    echo -e "${RED}[${severity}]${NC} $label — matches found (see above)"
+    WARNINGS=$((WARNINGS + 1))
+  fi
+}
+
+echo "::group::Obfuscation Scan"
+
+# ---------------------------------------------------------------------------
+# 1. OBFUSCATION PATTERNS
+# ---------------------------------------------------------------------------
+
+echo "→ Scanning for obfuscation patterns..."
+
+# Packed/obfuscated code blocks (long base64-like strings, hex chains)
+check_pattern \
+  "Packed/obfuscated JavaScript blob (self-decoding function pattern)" \
+  '(\w+),\s*\w+\s*=\s*function\s*\(\w+,\s*\w+,\s*\w+\)' \
+  "HIGH"
+
+# Large base64 strings (potential hidden payloads)
+check_pattern \
+  "Large base64-encoded string (≥100 chars, potential hidden payload)" \
+  '['\''"][A-Za-z0-9+/=]{100,}['\''"]' \
+  "HIGH"
+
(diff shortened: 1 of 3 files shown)


Lum1104:
@codex review this

tirth8205:
Reviewed with a focus on whether the scanner is sound, its false-positive/false-negative profile, and whether it's truly advisory vs blocking. The direction is good and the advisory-only framing is the right call, but the primary signal misfires in two places worth fixing before merge.

**Significant**

1. **The warning comment will 403 on fork PRs** (`supply-chain-scan.yml`). The workflow uses the `pull_request` trigger with no `permissions:` block. For fork PRs, GitHub scopes `GITHUB_TOKEN` to read-only, so the `actions/github-script` `issues.createComment` call in the "Flag for manual review" step will fail with 403 and post nothing. The job will still go red (the scanner exits 1), but the explanatory comment — described in the PR as the main signal — won't appear on exactly the fork PRs you call out as the highest-risk vector. To comment on fork PRs you need `pull_request_target` (or a `workflow_run` companion) with `issues: write`/`pull-requests: write`, being careful not to check out/execute untrusted code in the privileged context.

2. **`package.json` in `CONFIG_FILES` will fire on nearly every release PR** (`scan-obfuscation.sh`). The file-targeting check increments `WARNINGS` (→ exit 1 → comment) for any PR touching `package.json`. I tested it against a plain version-bump diff and it flags. Given that releases here bump the version across several manifest files, most release PRs will trip the scanner — that's alert fatigue that trains people to ignore the check. Consider limiting to auto-exec configs (astro/vite/next/webpack) or only flagging `package.json` when `scripts`/`dependencies` actually change.

**Minor**

3. `['"][A-Za-z0-9+/=]{100,}['"]` (base64 heuristic) matches data: URIs and any long quoted token (verified by test). `paths-ignore` skips svg/png/md but not `*.min.js`, dist bundles, or lockfiles, so inlined/minified content will produce false positives. (`sha512-` integrity hashes happen to be safe because the hyphen breaks the run.)

4. The job `if:` condition `…fork || github.event_name != 'pull_request_target'` is always true under the `pull_request` trigger — the second operand can never be false here. The condition and its inline comment are dead/misleading.

5. The global-assignment loop uses `grep -l` against a single file, so `$match` is always `/tmp/pr-diff.txt`. The message "Suspicious global assignment in build/config file: /tmp/pr-diff.txt" never names the real file, and the diff is the whole PR (not config-scoped), so "in build/config file" is inaccurate.

**Nits**
- `edited` in the trigger types re-runs the full scan on title/body edits, which don't change the diff. `opened, synchronize, reopened` is enough.
- Both new files are missing a trailing newline.

**Question**
- `git diff ${base.sha}..${head.sha}` after `checkout` of the fork head: is `base.sha` guaranteed to be in the fetched object graph for fork PRs even with `fetch-depth: 0`? If not, the diff step could fail with `fatal: bad object` instead of scanning. Worth confirming against a real fork PR.

**Overall**: The SECURITY.md changes are accurate and clear, and keeping the scan non-blocking (red check + soft comment, not a required gate) is the correct design for a regex heuristic. The main thing holding it back is that the comment won't post on fork PRs as written, and the `package.json` rule will generate routine noise. Fix #1 and #2 and this is a worthwhile guardrail.

PR #473, opened 2026-06-18
adding "/reload-plugins"

cope:
## Summary
On each install, I see: "Installed understand-anything. Run /reload-plugins to apply." - so it should be listed for users not to forget.

## How I tested this

No code changed.

## Versioning

- [X] N/A — internal/docs-only change


Lum1104:
Good catch, can you add to the translated version of README in `READMEs/` too?

cope:
sure, lemme check

cope:
@Lum1104 done.

PR #490, opened 2026-06-21
perf+robustness: eight verified hot-path fixes for large graphs (tour-gen O(n²)→O(n), single-parse, louvain crash, …)

DashBot-0001:
## What

Eight small, **behavior-preserving** fixes on hot paths, motivated by large-graph scaling (the 15k-file repo in #226, many-service monorepos, and the ~3k-node dashboard target). Each is verified against the project's existing tests (green before *and* after) and the perf/robustness-sensitive ones are benchmarked with byte-identical output.

| # | File | Change | Evidence |
|---|------|--------|----------|
| 1 | `skills/understand/merge-batch-graphs.py` | Precompile the 24-alternative project-prefix regex once at import instead of rebuilding + `re.escape`-ing the pattern *string* on every node. | `test_merge_batch_graphs.py` 69/69 pass; **19.5×** faster, 0 parity mismatches over 50k node ids. |
| 2 | `packages/core/src/analyzer/tour-generator.ts` | Replace `O(n²)` `topoOrder.includes()` (per code node) with a `Set`, and the `O(n)` `queue.shift()` Kahn dequeue with a head cursor. | `tour-generator.test.ts` green; **81×** faster at 20k nodes, byte-identical topo order. |
| 3 | `packages/core/src/analyzer/layer-detector.ts` | Collapse two full passes over `graph.nodes` into one; path-less file nodes deferred to `Core` to preserve ordering + Map key-insertion order. | `layer-detector.test.ts` green. |
| 4 | `packages/core/src/embedding-search.ts` | Hoist the query vector's magnitude out of the per-node cosine loop (invariant across a search). Same arithmetic, same order → bit-identical scores. | `embedding-search.test.ts` green. |
| 5 | `packages/dashboard/src/utils/louvain.ts` | Replace `Math.max(...Array.from(map.values())…)` with a reducing loop — the spread throws `RangeError: Maximum call stack size exceeded` at the ~3k+ node target. | byte-identical over 5000 fuzz + edge cases; old form throws at 1M ids, new doesn't. |
| 6 | `core/.../tree-sitter-plugin.ts`, `registry.ts`, `types.ts`, `skills/understand/extract-structure.mjs` | `extract-structure.mjs` parsed every code file **twice** (`analyzeFile` then `extractCallGraph`). Add an optional `analyzeFileFull()` that parses once and runs both extractors on the same tree (caller falls back when absent). Also cache one reusable tree-sitter parser per language. | `tree-sitter-plugin`/`plugin-registry`/`parsers` suites green before+after (132/132); against the real TS grammar `analyzeFileFull` ≡ `analyzeFile`+`extractCallGraph` (0 mismatches) at **~39% less parse work**/code file. |
| 7 | `skills/understand/compute-batches.mjs` | `buildNonCodeBatches` re-filtered the full path list once per Dockerfile/migration dir — `O(dirs·N)`. Hoist the path list once + index paths by dir; Group A/D lookups become `O(1)`. | byte-identical on the `scan-result-non-code.json` fixture + 300 fuzzed repos (0 mismatches); **17–31×** faster on 300–600-service monorepos (4250 files: 105ms → 3.4ms). |
| 8 | `packages/dashboard/src/utils/filters.ts` | `getEdgeCategory` linear-scanned every category's type array per edge in `filterEdges`. Build a reverse `edgeType → category` index once at module load; lookup is O(1). | byte-identical over all known edge types + unknowns (0 mismatches). |

## Why

`tour-generator`'s `includes()` is the clearest asymptotic trap on a 15k-file repo. #6 halves tree-sitter parse work for code files on the indexing hot path; #7 is the same shape for non-code files on many-service monorepos (the microservices-demo case). #5 removes a real crash vector at the dashboard's stated scale. The rest are free — same outputs, less work, all on paths that run on every `/understand`, every search, and every layer/tour build.

## Safety

All eight preserve behavior exactly. #2/#4/#5/#6/#7 produce byte-identical output (verified by diffing old vs new over large/fuzzed inputs and against the real tree-sitter grammar); #1/#3 are covered by the existing suites. #6 is additive (`analyzeFileFull` is optional, caller falls back). No breaking API changes.

## Related

- **#491** (filed alongside this) — the knowledge-graph view still runs d3-force **synchronously** on the main thread (~4.3s block at 3k nodes), and `dashboard/utils/layout.worker.ts` is dead code. That's a bigger, behavior-changing fix (worker-ization + deterministic seeding) so it's tracked separately rather than bundled here.
- `skills/understand-domain/extract-domain-context.py` — the gitignore glob→regex is unanchored (`build` matches `mybuild/`, over-ignoring); a correct fix needs proper segment-anchoring + a gitignore test suite, so it's left for a dedicated PR.


Lum1104:
Thanks — reviewed all eight changes and verified the equivalence claims independently (including re-running `analyzeFileFull` vs `analyzeFile` + `extractCallGraph` over 200 real files from this repo with the real grammars: 0 mismatches). Solid work.

Pushed three commits onto your branch (maintainer edits) so we can land this as-is:

1. **Merge of current `main`** — the branch was ~92 commits behind; the merge is clean and all suites pass on the merged result (core 931, skill 257, pytest 80, dashboard build, lint).
2. **`fix: return fresh arrays from analyzeFileFull degraded paths; add tests`** — the shared `EMPTY_STRUCTURE` static leaked the same array instances into every degraded-path result (`{ ...spread }` only copies the object, not the nested arrays), so one caller mutating its result would corrupt every later caller's. Replaced with a helper returning fresh arrays, and added the unit coverage the new API was missing (output equality, cached-parser stability, unsupported-extension fallback, mutation isolation, registry delegation/fallback).
3. **Version bump 2.9.1 → 2.9.2** per repo convention.

Will merge once CI is green. The #491 worker-ization is welcome as a follow-up — the deterministic-seed change you offered sounds like the right first step.

PR #499, opened 2026-06-22
fix(install): update kimi skill path to .agents/skills per-skill

wirechen:
Kimi Code CLI currently discovers user skills under `~/.kimi-code/skills/` and `~/.agents/skills/`, and only scans direct subdirectories (it does not recursively load nested skill directories).

The previous installer mapping for `kimi` used the legacy `~/.kimi/skills/` path with a folder symlink, so none of the `/understand` skills were loaded.

This PR updates the kimi platform entry to use `~/.agents/skills` with per-skill symlinks, matching the current official Kimi Code CLI behavior.

Fixes the issue where `/understand` is unavailable after installing for kimi.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -38,7 +38,7 @@ vibe|$HOME/.vibe/skills|per-skill
 vscode|$HOME/.copilot/skills|per-skill
 hermes|$HOME/.hermes/skills|folder
 cline|$HOME/.cline/skills|folder
-kimi|$HOME/.kimi/skills|folder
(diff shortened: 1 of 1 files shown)


Kenny-Huang-mz:
Yes, you're absolutely right, I still met this problem yesterday, it seems that they didn't merge your pull?

wirechen:
> Yes, you're absolutely right, I still met this problem yesterday, it seems that they didn't merge your pull?

没有，自己搞吧。其实这玩意儿并没有多好用

PR #513, opened 2026-06-25
fix: preserve real prefix when middle ID segment is a reserved word

JOhnsonKC201:
## Problem

`stripToValidPrefix` in `analyzer/normalize-graph.ts` collapses any node ID whose **second** segment happens to be a valid prefix, treating it as a double-prefix duplicate. This corrupts IDs where a reserved word legitimately appears as a middle path segment.

For example, `endpoint:service:getUser` is parsed as:
- outer segment `endpoint` (valid prefix) ✓
- next segment `service` (also a valid prefix) → wrongly assumed to be a duplicate prefix

…so the real `endpoint` prefix is dropped and the function returns `{ prefix: "service", path: "getUser" }`, yielding `service:getUser`. This:

- **changes the node type** (endpoint → service),
- **breaks edge references** that point at the original ID, and
- **violates idempotency** — normalizing an already-normalized ID mutates it.

## Fix

Only collapse a **true same-prefix duplicate** (e.g. `file:file:src/foo.ts`) by requiring the inner segment to equal the outer prefix:

```ts
// before
if (innerColonIdx > 0 && VALID_PREFIXES.has(rest.slice(0, innerColonIdx))) {

// after
if (innerColonIdx > 0 && rest.slice(0, innerColonIdx) === segment) {
```

A different reserved word in the middle is a legitimate path segment and is preserved. The genuine `file:file:...` double-prefix case still collapses as before.

## Tests

Added two regression tests to `normalize-graph.test.ts`:
- `endpoint:service:getUser` is preserved unchanged (was previously corrupted to `service:getUser`).
- normalization is idempotent for IDs with a reserved-word middle segment.

`pnpm --filter @understand-anything/core test` — **755 passing** (including the existing `file:file:` double-prefix test, which still passes).

### DIFF
--- a/understand-anything-plugin/packages/core/src/__tests__/normalize-graph.test.ts
+++ b/understand-anything-plugin/packages/core/src/__tests__/normalize-graph.test.ts
@@ -512,6 +512,49 @@ describe("normalizeBatchOutput", () => {
     expect(result.edges[0].target).toBe("file:src/foo.ts");
     expect(result.stats.danglingEdgesDropped).toBe(0);
   });
+
+  it("repairs an edge endpoint with a chain of reserved-word prefixes", () => {
+    // Regression: an edge endpoint "service:endpoint:file:src/foo.ts" carries
+    // more than one reserved prefix before the real "file" prefix. Normalizing
+    // the full id for each candidate type can't collapse the run, so the repair
+    // must normalize from the candidate prefix segment to resolve it to the
+    // canonical node "file:src/foo.ts" rather than drop the edge.
+    const result = normalizeBatchOutput({
+      nodes: [
+        {
+          id: "file:src/foo.ts",
+          type: "file",
+          name: "foo.ts",
+          filePath: "src/foo.ts",
+          summary: "Target",
+          tags: [],
+          complexity: "simple",
+        },
+        {
+          id: "file:src/bar.ts",
+          type: "file",
+          name: "bar.ts",
+          filePath: "src/bar.ts",
+          summary: "Source",
+          tags: [],
+          complexity: "simple",
+        },
+      ],
+      edges: [
+        {
+          source: "file:src/bar.ts",
+          target: "service:endpoint:file:src/foo.ts",
+          type: "imports",
+          direction: "forward",
+          weight: 0.7,
+        },
+      ],
+    });
+
+    expect(result.edges).toHaveLength(1);
+    expect(result.edges[0].target).toBe("file:src/foo.ts");
+    expect(result.stats.danglingEdgesDropped).toBe(0);
+  });
 });
 
 describe("normalizeBatchOutput integration", () => {
--- a/understand-anything-plugin/packages/core/src/__tests__/normalize-graph.test.ts
+++ b/understand-anything-plugin/packages/core/src/__tests__/normalize-graph.test.ts
@@ -555,6 +555,54 @@ describe("normalizeBatchOutput", () => {
     expect(result.edges[0].target).toBe("file:src/foo.ts");
     expect(result.stats.danglingEdgesDropped).toBe(0);
   });
+
+  it("drops a missing canonical endpoint instead of rewiring it to a same-named node", () => {
+    // Regression: an edge endpoint "func:src/a.ts:formatDate" is a canonical
+    // multi-part ID whose node is missing, while a DIFFERENT function
+    // "func:formatDate" exists. The reserved-prefix peeling must not strip the
+    // real "func" prefix and normalize the bare suffix "src/a.ts:formatDate"
+    // back to "func:formatDate" — that silently rewires the edge to an
+    // unrelated function. The edge must be dropped as dangling instead.
+    const result = normalizeBatchOutput({
+      nodes: [
+        {
+          id: "func:formatDate",
+          type: "function",
+          name: "formatDate",
+          summary: "A different formatDate",
+          tags: [],
+          complexity: "simple",
+        },
+        {
+          id: "file:src/a.ts",
+          type: "file",
+          name: "a.ts",
+          filePath: "src/a.ts",
+          summary: "Source",
+          tags: [],
+          complexity: "simple",
+        },
+      ],
+      edges: [
+        {
+          source: "file:src/a.ts",
+          target: "func:src/a.ts:formatDate",
+          type: "calls",
+          direction: "forward",
+          weight: 0.7,
+        },
+      ],
+    });
+
+    expect(result.edges).toHaveLength(0);
+    expect(result.stats.danglingEdgesDropped).toBe(1);
+    expect(result.stats.droppedEdges[0]).toEqual({
+      source: "file:src/a.ts",
+      target: "func:src/a.ts:formatDate",
+      type: "calls",
+      reason: "missing-target",
+    });
+  });
 });
 
 describe("normalizeBatchOutput integration", () => {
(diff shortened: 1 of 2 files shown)


JOhnsonKC201:
Confirmed the codex P2 and pushed a follow-up. The earlier `inner === segment` check kept a spurious reserved-word project prefix — e.g. `service:file:src/foo.ts` for a `file` node stayed as-is instead of resolving to the canonical `file:src/foo.ts`, which would dangle edges referencing the canonical ID.

Fix (6d54111): `stripToValidPrefix` now takes the node's expected prefix and collapses the outer segment when the inner one is **either** the same reserved word (true duplicate) **or** the expected prefix (spurious reserved-word project prefix). A different reserved word that isn't the expected prefix (`endpoint:service:x`) is still preserved as a real path segment. Added two regression tests (strip + idempotency); 757/757 core tests pass.

@codex review

JOhnsonKC201:
Addressed the edge-path P2 from the last review. The node-level fix didn't cover `normalizeBatchOutput`'s edge-repair fallback: for an endpoint like `service:file:src/foo.ts` pointing at canonical node `file:src/foo.ts`, `inferTypeFromId` read the spurious `service` as the type, so the strip never fired and the edge dropped as dangling.

Fix (0a0ffd2): added `resolveEdgeEndpoint` — tries the prefix-inferred type first (common case unchanged), then each subsequent leading reserved-word segment as a candidate type, returning the first that resolves to an existing node (and the original id when none do). Added a regression test asserting the edge is repaired rather than dropped; 758/758 core tests pass.

@codex review

JOhnsonKC201:
Addressed the third codex P2 ("Normalize from the candidate prefix segment"). The edge-repair fallback normalized every candidate type against the *full* endpoint id, so a chain of reserved prefixes before the real one — e.g. `service:endpoint:file:src/foo.ts` pointing at `file:src/foo.ts` — couldn't be collapsed for the `file` candidate and the edge was dropped as dangling.

`resolveEdgeEndpoint` now pairs each candidate type with the id suffix that begins at its prefix segment and normalizes from that suffix. The inferred-type/full-id attempt is still tried first, so the common case and the earlier single reserved-word prefix case are unchanged. Added a regression test for the multi-prefix chain; 759/759 core tests pass.

JOhnsonKC201:
@codex review

Mars-Wave:
you don't want to follow up on that last codex review? @codex reprimand

JOhnsonKC201:
@Mars-Wave sorry for the long silence here. I did follow up on that review, but I pushed the commit without saying anything, so from the outside this has looked abandoned for a month. That is on me.

The fourth P2 ("Avoid stripping real prefixes during edge repair", `normalize-graph.ts:264`) is fixed in `3271413`, which is the current head of this branch.

The problem was that the edge-repair fallback peeled *every* leading reserved-word segment, including an endpoint's own real prefix. A missing `func:src/a.ts:formatDate` had `func` stripped, the bare remainder `src/a.ts:formatDate` re-normalized, and the result resolved to an unrelated `func:formatDate` node. So instead of dropping as dangling, the edge was silently rewired to a different function with the same name, which is worse than the dangling edge it was trying to avoid.

The fix only peels a leading reserved word when what remains is itself a canonical `prefix:path` id, which is the signature of a spurious outer prefix sitting in front of the real one. When the remainder is a bare path, the peeled segment *was* the endpoint's own prefix, so it is left intact and the missing endpoint drops as dangling:

```ts
rest = rest.slice(colonIdx + 1);
// Only strip `segment` as a spurious outer prefix when the remainder is
// still a canonical prefixed id. Otherwise `segment` was the real prefix
// and the bare suffix must not be reinterpreted as another node.
if (!startsWithValidPrefix(rest)) break;
```

Added a regression test covering exactly that case: a missing `func:src/a.ts:formatDate` endpoint with a decoy `func:formatDate` node present, asserting the edge drops rather than rewires. The earlier multi-prefix chain case (`service:endpoint:file:src/foo.ts`) still resolves as before, so the fix is narrow.

760/760 core tests pass on the branch.

@codex review

JOhnsonKC201:
Good catch, and it is a regression I introduced rather than a pre-existing gap, so thanks for running it again.

`6d54111` narrowed the node-side collapse to fire only when the immediate inner segment is the same reserved word or the node's expected prefix. That fixed the original over-collapse, but it stopped handling a *run* of more than one spurious prefix. For a file node emitted as `service:endpoint:file:src/foo.ts` the inner segment is `endpoint`, not `file`, so it returned at `service` and left the ID uncanonical, dangling every edge pointing at `file:src/foo.ts`. The normalizer before this PR collapsed that run, so it was a step backwards for that input.

Worth noting why the existing test did not catch it: the chain case added in `67c6f19` exercises `service:endpoint:file:src/foo.ts` as an *edge endpoint*, which goes through `resolveEdgeEndpoint`, not as the node ID itself. Exactly the gap you described.

Fixed in `1c509ca`. The collapse now also fires when the outer segment is not the expected prefix and the expected prefix is still ahead in the leading run of reserved words:

```ts
(expectedPrefix !== undefined &&
  segment !== expectedPrefix &&
  chainContainsPrefix(rest, expectedPrefix))
```

`chainContainsPrefix` walks that leading run and stops at the first segment that is not a reserved word, so ordinary path text that happens to contain the word cannot trigger it. The `segment !== expectedPrefix` guard is what keeps a real prefix intact, so both of these still pass through untouched on an endpoint node:

- `endpoint:service:getUser` (the original bug this PR opened on)
- `endpoint:file:handler` (a real prefix followed by a reserved word as a genuine path segment)

Three tests added: the chain collapsing whole, idempotency on the result, and that over-collapse guard. Only the first fails on the previous commit, which keeps the change confined to the regression.

763/763 core tests pass, tsc and eslint clean.

@codex review

JOhnsonKC201:
Went through the three review threads that are still open on `1c509ca`. Two are already fixed at head, the third reproduces but is pre-existing rather than something this PR introduces. Details, since two of these look identical from the diff:

**"Collapse reserved-prefix chains on nodes" (line 94): fixed.** This was a real regression I introduced in `6d54111` by narrowing the collapse condition, and `1c509ca` fixes it with `chainContainsPrefix`. `normalizeNodeId("service:endpoint:file:src/foo.ts", { type: "file" })` returns `file:src/foo.ts`, and it is idempotent. Covered by three tests, including a guard that `endpoint:file:handler` on an endpoint node is still left alone.

**"Avoid stripping real prefixes during edge repair" (line 294): fixed in `3271413`.** I checked it again rather than trusting the commit message. With `func:formatDate` present and `func:src/a.ts:formatDate` missing, the edge drops as dangling instead of rewiring. That is the behaviour the thread asked for.

**"Avoid peeling canonical prefixes with prefixed path segments" (line 294): reproduces, but it is not a regression.** With a `file:handler` node present, an edge to a missing `endpoint:file:handler` does resolve to `file:handler`. I ran the same case against `origin/main` and it produces exactly the same rewire, so this behaviour predates the PR and is unchanged by it.

I do not think it can be fixed on the edge path without new information, and I would rather say so than add another guess. `service:file:src/foo.ts` and `endpoint:file:handler` are the same shape, `<reserved>:<reserved>:<path>`. The first must be peeled, the second must not, and the only thing separating them is the node's declared type. The node path has that and uses it as `expectedPrefix`. An edge endpoint does not carry one. I also tried the obvious tiebreak, re-deriving the original id with the resolved node's type and requiring it to reproduce that node, and it confirms both cases, so it does not discriminate. Any guard strict enough to drop `endpoint:file:handler` also drops `service:file:src/foo.ts`, which is the case this PR exists to fix.

So my suggestion is to treat it as a separate issue. There is one signal that would work, which is skipping the ambiguous peel when the batch actually contains nodes of the peeled prefix's type, but that is a behaviour change to pre-existing code and does not belong in this PR. Happy to open it separately if you want it.

Otherwise this is ready from my side: 763 core tests green, tsc and eslint clean. Let me know if you would prefer I scope anything back.

Issue #340, opened 2026-05-31
[Windows] Codex fails to locate agents under ~/.understand-anything-plugin/agents/ unless full path is specified


kellydxy0307-cell:
**Platform**: Windows (Codex CLI)  
**Setup method**:  
`iwr -useb https://raw.githubusercontent.com/Lum1104/Understand-Anything/main/install.ps1 | iex`

**Observed layout after installation**:
- `~/.agents`
- `~/.understand-anything`
- `~/.understand-anything-plugin` (junction)

**Issue**:
When running:
“$understand-anything:understand --language en --full”
Codex fails to locate the agent markdown files.

**Hypothesis**:
Codex CLI does not automatically resolve the `~/.understand-anything-plugin` parent path when interpreting agent references on Windows

**Expected behavior**:
The default agent path should be recognized without manual specification.

PR #524, opened 2026-06-29
fix(skills): force explicit UTF-8 on locale-default Python file reads

H179922:
## Problem

Two Python files read with no explicit `encoding=`. `pathlib`'s `read_text`/`write_text` fall back to `locale.getpreferredencoding(False)` when none is given. On Windows — absent Python UTF-8 mode (PEP 540) — that's the ANSI codepage (`cp1252` / `cp936` / `cp932`), **not UTF-8**.

**`skills/understand-domain/extract-domain-context.py`** (lines 126, 211, 270, 315, 418) reads source files, `.gitignore`, and metadata, then writes `domain-context.json`. On Windows the reads decode UTF-8 source against the wrong codepage and silently corrupt the context that the `domain-analyzer` agent consumes — CJK comments, accented identifiers, em-dashes, the translated README text pulled in at line 332.

`errors="replace"` does **not** catch this: legacy codepages decode almost every byte to *some* (wrong) character instead of raising, so there's no `UnicodeDecodeError` and no replacement char — just silent mojibake.

### Reproduction

A UTF-8 source file with a CJK comment, read the way Windows would:

```
source (utf-8) : '# 用户订单服务 — order service'
cp1252 (EN Win): '# ç”¨æˆ·è®¢å•æœ�åŠ¡ â€”'
cp936  (CN Win): '# 鐢ㄦ埛璁㈠崟鏈嶅姟 鈥 order s'
```

Both alternate reads used `errors="replace"` — neither raised, both are silently wrong.

**`tests/skill/understand/test_merge_batch_graphs.py`** (lines 986, 1105) reads `assembled-graph.json` back for assertions. `merge-batch-graphs.py` writes it with `ensure_ascii=False`, so it can contain raw non-ASCII that the unencoded read would mojibake on Windows, failing the test there.

## Fix

Pass `encoding="utf-8"` to the affected reads/write. This matches the convention already used in every other Python script in the repo (`merge-batch-graphs.py`, `merge-subdomain-graphs.py`, `parse-knowledge-base.py`, `merge-knowledge-graph.py`). No behavior change on Linux/macOS, where the locale default is already UTF-8.

## Scope

7 lines across 2 files. Relates to the Windows-compat reports #262 and #340.

### DIFF
--- a/tests/skill/understand/test_merge_batch_graphs.py
+++ b/tests/skill/understand/test_merge_batch_graphs.py
@@ -983,7 +983,7 @@ class TestMultiPart(unittest.TestCase):
             capture_output=True, text=True,
         )
         out_path = self.intermediate / "assembled-graph.json"
-        assembled = _j.loads(out_path.read_text()) if out_path.exists() else {}
+        assembled = _j.loads(out_path.read_text(encoding="utf-8")) if out_path.exists() else {}
         return result.returncode, result.stderr, assembled
 
     def test_two_parts_of_one_logical_batch_merge(self) -> None:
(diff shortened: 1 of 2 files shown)


Lum1104:
@codex review this

PR #537, opened 2026-07-03
feat(core): add Swift tree-sitter support

KumamuKuma:
## Summary
- add a workspace-vendored tree-sitter Swift WASM package and wire `swiftConfig` into the tree-sitter loader
- add `SwiftExtractor` coverage for Swift type containers, functions/methods, visibility, imports/exports, and call graph extraction
- add Swift import-map resolution for SwiftPM modules, including `Package.swift` target paths, so Swift test targets can connect to source targets

## Scope
This PR is intentionally Swift-only. The Windows CI skill-test fix is kept separate in #538.

## CI note
Ubuntu passes on this branch. The current Windows failure is caused by existing Windows skill-test issues in `main` that are fixed by #538, not by the Swift extractor changes. Recommended order: merge #538 first, then update/rebase this PR so #537 can rerun on the fixed Windows CI baseline.

## Why
Swift files were previously scanned as code but had no tree-sitter grammar/extractor, so `.swift` files produced empty structural analysis and no useful graph nodes/edges. This addresses the Swift side of #226 and follows the workspace-vendored WASM pattern used after #435, avoiding the third-party bundle approach from #344.

## Local tests
- `corepack pnpm --filter @understand-anything/core build`
- `corepack pnpm --filter @understand-anything/core test`
- `corepack pnpm exec vitest run tests/skill/understand/test_extract_import_map.test.mjs tests/skill/understand/test_scan_project.test.mjs`


Lum1104:
Thanks for your PRs, merged them. Awesome!

PR #551, opened 2026-07-06
fix(skills): limit changed-file batches to changed targets

### LINKED ISSUE #547: compute-batches.mjs --changed-files re-analyzes whole communities, not just changed files (token waste + graph churn)
**Version:** understand-anything 2.7.5

### Summary
On incremental updates, `compute-batches.mjs --changed-files=<list>` emits batches whose `files[]` (the analysis targets) include **every file in any community that contains a changed file**, not just the changed files. The file-analyzer then re-analyzes many unchanged neighbors, which (a) wastes a large multiple of tokens and (b) churns the graph — LLM re-wording of unchanged nodes' summaries/tags produces noisy diffs on files that did not change.

### Repro
Large TS monorepo (~1,140 files). A refresh with **85 changed files** produced:
```
Wrote 18 batches ... 
distinct analysis-target files: 351
  in changed set: 85
  NOT in changed set: 266   <-- unchanged neighbors pulled in as full targets
```
So ~4× the necessary files get sent to the analyzer subagents.

### Expected
Per SKILL.md, the incremental path should analyze only the changed files, with unchanged neighbors present **only** in `neighborMap` (for cross-batch edge context), not in `files[]`:
> produces a batches.json that contains only batches with changed files, but neighborMap entries still reference unchanged files …

The current output instead promotes whole communities into `files[]`.

### Impact
Defeats much of the point of an incremental update. On a staleness refresh the token cost scales with community size rather than change size, and unchanged nodes get regenerated (diff churn), which is exactly what incremental mode is meant to avoid.

### Workaround
Post-process `batches.json` before dispatch: intersect each batch's `files[]` with the changed-set, drop now-empty batches, keep `neighborMap` and `batchImportData` intact. This drops targets from 351 → 85 with no loss of cross-batch edge fidelity (the merge's importMap-recovery + dangling-edge dropper still resolve edges to the reused baseline nodes).

### Suggested fix
In `--changed-files` mode, restrict each emitted batch's `files[]` to the changed set (files that actually need re-analysis), and rely on `neighborMap` for the surrounding community context — rather than emitting the full community as analysis targets.

KumamuKuma:
## Summary
- Limit `--changed-files` batches to changed analysis targets instead of full Louvain communities.
- Rebuild `batchImportData` and `neighborMap` for changed targets while preserving full-graph `batchIndex` context.
- Normalize Windows-style changed-file paths (CRLF, backslashes, leading `./`).

## Related Issue
- Fixes #547

## Test Plan
- `pnpm vitest run tests/skill/understand/test_compute_batches.test.mjs`
- `pnpm test`
- `pnpm build`
- `pnpm lint`

### DIFF
--- a/tests/skill/understand/test_compute_batches.test.mjs
+++ b/tests/skill/understand/test_compute_batches.test.mjs
@@ -577,20 +577,22 @@ describe('compute-batches.mjs — --changed-files', () => {
     if (root) rmSync(root, { recursive: true, force: true });
   });
 
-  it('emits only batches containing changed files', () => {
+  it('emits only changed files from retained batches with Windows-style changed-file paths', () => {
     root = setupProject('scan-result-3-cliques.json');
     const changedPath = join(root, 'changed.txt');
-    // Only the auth clique is changed
-    writeFileSync(changedPath, ['src/auth/login.ts', 'src/auth/tokens.ts'].join('\n'));
+    // Only two files in the auth clique are changed. Use CRLF plus one
+    // backslash path to cover Windows git diff/path-list inputs.
+    writeFileSync(changedPath, ['src\\auth\\login.ts', 'src/auth/tokens.ts'].join('\r\n'));
 
     const result = runScript(root, [`--changed-files=${changedPath}`]);
     expect(result.status).toBe(0);
 
     const out = readBatches(root);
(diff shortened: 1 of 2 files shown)


KumamuKuma:
@Lum1104 please help check, thank you very much!

Issue #413, opened 2026-06-07
Enhancement ideas from heavy use on a large, fast-moving Rust codebase


MFJG-dev:
Thanks for UA — we've been running it on a ~2000-file, 40-crate Rust audio workstation that lands ~30 PRs/day. The skills (especially `/understand-explain` and `/understand-diff`) have been genuinely useful. A few enhancements from that usage, roughly highest-leverage first:

## 1. Cheap incremental refresh keyed on a ref
`/understand --since <commit>` that re-analyzes only files changed since the graph's `meta.json` `gitCommitHash`, instead of a full rebuild. On a fast-moving repo the full rebuild cost is the main barrier to keeping the graph current — a refresh proportional to the actual drift would make 'rebuild periodically' nearly free. (Auto-update-on-commit is too aggressive at our velocity; an explicit incremental command is the sweet spot.)

## 2. Pre-flight cost/scope estimate
Phase 1 (`scan-project.mjs`) already computes file count + complexity ('very-large') + batch count. Surfacing an estimated token/time cost *before* `/understand --full` starts would let users decide to proceed or scope down. The data's already gathered; it's just not shown up front.

## 3. Staleness is silent
`meta.json` records `gitCommitHash` + `lastAnalyzedAt`, but nothing warns when the graph is far behind HEAD — ours sat ~650 commits stale before we noticed, and graph-dependent skills (`/understand-chat`, dashboard) answered confidently from the stale graph. A freshness check (commits-behind / days-behind, or fingerprint-mismatch count using the existing `fingerprints.json`) that warns before trusting an old graph would prevent confidently-wrong output.

## 4. A committable prose artifact, separate from the heavy graph
`explain-notes.md` was the one output worth versioning in-repo (it ages gracefully), but it sits next to the 50+ MB of graph/fingerprint JSON that must stay gitignored. A `--emit-notes` that writes just a small architecture summary, distinct from the bulk graph, would let teams commit the durable signal without the rot-prone bulk.

## 5. Rust language addendum
`languages/` has no `rust.md` and `frameworks/` is web-only, so Rust projects get only the base template. A `languages/rust.md` (crates/workspaces, traits, feature gates, `unsafe`/FFI boundaries, async runtimes) would sharpen summaries considerably for the Rust ecosystem.

Happy to contribute PRs for any of these if they're directions you'd welcome — particularly the Rust addendum (5) and the staleness check (3), which are self-contained. Thanks again!

[...]

PR #558, opened 2026-07-08
feat: warn when knowledge graphs are stale

jiang4wqy:
## Why

Understand Anything graphs are snapshots. After source files or Git history change, the dashboard and graph-consuming skills could continue presenting old relationships without telling the user. This addresses item 3 of #413: graph staleness was silent.

## What the initial PR version missed

The first version established the warning path, but its freshness model was too narrow for real repositories:

- it checked only once when the dashboard mounted, so an already-open window did not update after later edits or commits
- it compared only the embedded graph commit with `HEAD`, so staged, unstaged, and untracked source changes were invisible
- it treated commit mismatches as a generic stale/behind state instead of distinguishing `behind`, `ahead`, and `diverged` histories
- repository-wide checks could report changes from a sibling project in a monorepo
- it checked only the knowledge graph, not the optional domain graph independently
- synchronous Git execution could block the dashboard server, and there were no explicit timeout/output limits
- the client trusted the endpoint shape too loosely, while errors and cache behavior were not hardened
- graph-consuming skills passed the stored hash directly to `git diff` without first resolving it as a commit
- the static demo could attempt a freshness request even though no live Git backend exists

### Core freshness engine

- adds asynchronous, shell-free Git execution with a 5-second timeout and bounded output
- reports `fresh`, `dirty`, `stale`, or `unknown`, with structured reasons instead of guessing
- detects staged, unstaged, and untracked project changes using NUL-delimited paths
- scopes commit and worktree checks to the selected project and ignores generated `.understand-anything/` output
- resolves abbreviated graph hashes safely and classifies `behind`, `ahead`, and `diverged` histories
- evaluates knowledge and domain graph commits from one repository snapshot
- preserves the legacy staleness exports for compatibility

### Dashboard behavior

- exposes token-protected `/staleness.json` data with sanitized failures and `Cache-Control: no-store`
- validates the complete response shape before rendering it
- fetches once on load, then refreshes only when the browser window receives `focus`
- aborts superseded requests and ignores late responses, preventing stale network results from overwriting newer state
- shows graph-specific status, relation-aware guidance, and up to eight changed paths on desktop and mobile
- skips freshness requests in the static demo unless `VITE_STALENESS_URL` is explicitly configured

### Skill preflight and release metadata

- updates `understand-chat`, `understand-diff`, `understand-domain`, `understand-explain`, and `understand-onboard`
- resolves the stored graph commit with `git rev-parse --verify --end-of-options` before using it
- emits a soft warning without blocking the normal skill workflow
- bumps all five plugin manifests from `2.8.2` to `2.8.3`

## Scope

This PR detects and explains stale graphs; it does not silently regenerate them. Regeneration remains an explicit `/understand` or `/understand --force` action so opening or refocusing the dashboard cannot trigger expensive analysis unexpectedly.

## Related work

- Addresses item 3 of #413 (silent graph staleness). #413 is an umbrella issue, so this PR intentionally does not auto-close it.
- Related context: #152 tracks fingerprint merge corruption. This PR does not modify the graph merge algorithm.
- Partial implementation overlap: #452 fixes NUL-delimited paths in the legacy incremental merge helper. This PR independently uses NUL parsing in the new asynchronous freshness engine, so merge order should preserve #452's legacy-path fix.

## Validation

Final publish branch after merging current `main`:

- `pnpm lint`
- Core build: `pnpm --filter @understand-anything/core build`
- Skill build: `pnpm --filter @understand-anything/skill build`
- Core tests: 41 files, 911 tests passed
- Dashboard tests: 10 files, 80 tests passed
- Root tests: 22 files passed, 1 conditionally skipped; 267 tests passed, 6 conditionally skipped
- Dashboard production build and static demo build passed
- Full workspace `pnpm build` passed
- `git diff --check origin/main..HEAD` passed
- production dashboard bundle contains no `child_process` runtime import

Manual browser validation used a real repository graph with 432 tracked files, 97 nodes, 183 edges, and 7 layers. Dirty, stale, and fresh transitions updated on window focus without reload; the changed-path summary and 390 px mobile layout were also checked.


jiang4wqy:
@ZebangCheng

ZebangCheng:
Thank you @jiang4wqy

PR #569, opened 2026-07-12
feat: native OpenClaw gateway plugin (in-process analysis pipeline + dashboard route)

zeroaltitude:
## Summary

image

Adds `.openclaw-plugin/` — a real [OpenClaw](https://github.com/openclaw/openclaw) gateway plugin that runs the Understand-Anything knowledge-graph pipeline **inside the gateway process itself**, as a deeper alternative to the existing `install.sh openclaw` skill-symlink integration (which is untouched and keeps working).

- **Agent tools** available to any connected agent: `understand_list_projects`, `understand_analyze_project` (background job: tree-sitter structural pass + one LLM call per file, persisted to `.ua/knowledge-graph.json`), `understand_status`, `understand_search` (via `@understand-anything/core`'s `SearchEngine`), `understand_get_node`.
- **HTTP routes** mounted on the gateway: `/understand-anything` (project picker) and `/understand-anything/open?project=<idx>`, which starts (or reuses) an `understand-anything-viewer` instance per project and redirects into its token-protected, 127.0.0.1-only dashboard — reusing the existing viewer's security model as-is.
- The pipeline (`src/pipeline.ts`) is built entirely on `@understand-anything/core`'s already-public API — `createIgnoreFilter`/`LanguageRegistry` for the walk, `TreeSitterPlugin.analyzeFile` for structure, `buildFileAnalysisPrompt`/`parseFileAnalysisResponse` + `buildProjectSummaryPrompt`/`parseProjectSummaryResponse` for LLM enrichment, `GraphBuilder` → `validateGraph` → `saveGraph`/`saveMeta` for persistence. No new parsing/prompt logic — just new orchestration and a new host.
- Output is byte-compatible with the rest of the ecosystem: same `.ua/knowledge-graph.json` shape the skills produce, same dashboard renders it.
- Analysis and serving are restricted to a configured `projects` allowlist (required, no default). Anthropic API key comes from plugin config or the gateway process env only — never written to disk by the plugin.
- `.openclaw-plugin` joins the pnpm workspace (`workspace:*` on `@understand-anything/core` and `understand-anything-viewer`) so it builds against the existing packages with no duplication.

## Why

OpenClaw's current support here is a skill symlink (`install.sh openclaw` → `~/.openclaw/skills/`), same mechanism as every other CLI platform. That's great for the Claude-Code-style slash-command flow, but it means the analysis pipeline only runs when a Task-tool-capable coding agent drives it step-by-step through the 848-line `/understand` skill. This plugin instead runs the pipeline as ordinary backend code attached to the gateway's own lifecycle — any agent connected to that gateway gets `understand_*` tools mid-conversation and a persistent dashboard route, with no CLI-specific dispatch required.

## Test plan

- [x] `pnpm --filter @understand-anything/openclaw-plugin build` (tsc) — clean
- [x] `pnpm --filter @understand-anything/openclaw-plugin test` — 13/13 unit tests pass (import-resolution + glob-matching)
- [x] Live pipeline run against a real ~50-file TypeScript project using a real Anthropic key: N nodes/edges produced, per-function LLM summaries accurate, import edges correctly resolved (including the TS ESM `./util.js` → `util.ts` case), 0 warnings
- [x] Persisted `.ua/knowledge-graph.json` validates against the existing schema; file paths correctly relativized (no absolute-path leaks)
- [x] `understand-anything-viewer` serves the resulting graph: dashboard loads, graph JSON accessible with token, `403` without
- [x] Loaded into a real OpenClaw gateway runtime (`openclaw plugins inspect <id> --runtime`) in an isolated config: plugin status `loaded`, all 5 tools + 2 HTTP routes registered, 0 error diagnostics
- [x] Ran live against this fork's sibling `openclaw-plugins` monorepo through an actual gateway + chat session — analysis completed successfully end-to-end (52/52 files) via a real tool call from an agent conversation

## Notes for reviewers

- This PR contains only the plugin itself — no changes to any existing skill, agent prompt, or core package.
- The commit history here is a clean cherry-pick of the actual plugin work onto current `main`; an earlier iteration of the branch had unrelated repo-tooling commits (a local `bd` issue-tracker init) mixed in — those are intentionally excluded from this PR.

### DIFF
--- a/.openclaw-plugin/README.md
+++ b/.openclaw-plugin/README.md
@@ -23,9 +23,21 @@ connected agent can call mid-conversation, plus a dashboard route.
 **HTTP routes** (mounted on the gateway):
 
 - `GET /understand-anything` — project picker
-- `GET /understand-anything/open?project=<idx>` — starts (or reuses) a
-  `understand-anything-viewer` instance for that project and redirects to its
-  token-protected dashboard URL. The viewer binds to 127.0.0.1 only.
+- `GET /understand-anything/open?project=<idx>` — starts (or reuses) a viewer
+  instance for that project and redirects to its token-protected dashboard
+  URL. Binds to 127.0.0.1 only.
+
+  With an Anthropic API key configured, this serves **interactive-server.ts**
+  instead of the plain upstream viewer: the identical dashboard + read-only
+  JSON API, plus a floating "Ask" chat panel backed by a live LLM
+  (`POST /ask.json`) — grounded in the persisted knowledge graph via
+  `SearchEngine`, the same idea as upstream's `/understand-chat` skill, just
+  reachable from the browser instead of a CLI. With no key configured, it
+  falls back to the plain zero-LLM `understand-anything-viewer` unchanged.
+  This split is deliberate: the standalone viewer's whole reason to exist is
+  staying LLM-free for team-sharing, so that path is never modified — the
+  interactive server is an additive, separate script that happens to reuse
+  its static assets.
 
 ## How it works
 
@@ -40,6 +52,16 @@ Messages API (`src/llm.ts`) — bounded by `concurrency` and `maxFiles`.
 
 Output is byte-compatible with the rest of the ecosystem: the same
 `.ua/knowledge-graph.json` the skills produce, and the same dashboard renders.
+
+The Ask panel (`src/ask.ts`) works the same way at query time: `SearchEngine`
+finds the graph nodes most relevant to the question, a bounded number of
+their source files are read for grounding, and a single LLM call (the same
+`llm.ts` caller used for analysis) answers using only that context — no
+re-scanning the project per question. `src/interactive-server.ts` is a
+distinct server from `understand-anything-viewer` (same static dashboard
+build + JSON API, reused read-only) that adds the `/ask.json` endpoint and
+injects a small vanilla-JS chat widget (`src/ask-widget.js`, no build step)
+into the served `index.html`.
 
 ## Install
 
@@ -81,6 +103,10 @@ understand_status
 understand_search { "query": "session management" }
 ```
 
+Open `/understand-anything` in a browser to view the dashboard. With an API
+key configured (as above), you'll see a floating chat button — ask it
+anything about the analyzed codebase and it answers grounded in the graph.
+
 ## Security notes
 
 - Analysis and serving are restricted to the configured `projects` allowlist.
@@ -88,4 +114,10 @@ understand_search { "query": "session management" }
   per-instance random access token, graph-derived file allowlist, 1 MB/no-binary
   caps on source preview.
 - The API key is only read from plugin config or the gateway process env; it is
-  never written to disk by the plugin.
+  never written to disk by the plugin. It's passed to the interactive server
+  subprocess via an environment variable, never a CLI arg (CLI args are
+  visible in `ps`; env vars of a process you own are not).
+- `/ask.json` requires the same per-instance access token as the graph/file
+  endpoints (sent as `X-Ask-Token` instead of a query param, since it's a
+  POST body, not a GET link) and only ever reads files already listed in the
+  persisted knowledge graph, same as `/file-content.json`.
--- a/.openclaw-plugin/README.md
+++ b/.openclaw-plugin/README.md
@@ -39,6 +39,13 @@ connected agent can call mid-conversation, plus a dashboard route.
   interactive server is an additive, separate script that happens to reuse
   its static assets.
 
+  Unanalyzed projects get an **"Understand this project" button** right on
+  this page — no need to go find the tool call. With `allowAddProject: true`
+  (see Install below), the picker also gets an **"Add a project"** form: paste
+  a `https://github.com/owner/repo` URL (shallow-cloned into
+  `~/.local/share/understand-anything-plugin/clones`) or an existing local
+  path, and it's added to the list and analysis starts immediately.
+
 ## How it works
 
 The pipeline (`src/pipeline.ts`) is built entirely on `@understand-anything/core`'s
@@ -62,6 +69,17 @@ distinct server from `understand-anything-viewer` (same static dashboard
 build + JSON API, reused read-only) that adds the `/ask.json` endpoint and
 injects a small vanilla-JS chat widget (`src/ask-widget.js`, no build step)
 into the served `index.html`.
+
+Project registration (`src/project-store.ts`) keeps config-declared projects
+(fixed, from `projects` below) and dynamically-added ones (persisted to
+`dynamic-projects.json` in the same state dir, so they survive a restart) in
+one combined, index-stable list — new entries only ever append. Adding a
+GitHub URL only recognizes an actual `https://github.com/...` or
+`git@github.com:...` URL (never a bare `owner/repo` shorthand, `file://`, or
+non-GitHub git URL — those could otherwise be used to smuggle an arbitrary
+local path or host into a "clone", defeating the projects allowlist this
+whole feature sits inside) and shells out to `git clone --depth 1` with a
+3-minute timeout.
 
 ## Install
 
@@ -87,7 +105,8 @@ Then in your `~/.openclaw/openclaw.json`:
           "model": "claude-sonnet-5",        // optional
           "concurrency": 5,                   // optional
           "maxFiles": 400,                    // optional
-          "anthropicApiKey": "sk-ant-..."    // optional; falls back to ANTHROPIC_API_KEY on the gateway process
+          "anthropicApiKey": "sk-ant-...",   // optional; falls back to ANTHROPIC_API_KEY on the gateway process
+          "allowAddProject": true             // optional, default false — lets the dashboard register new projects at runtime
         }
       }
     }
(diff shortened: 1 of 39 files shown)


zeroaltitude:
Ran an independent adversarial code review of this PR (GPT-5.6-Sol) and pushed a follow-up commit addressing everything at must-fix/should-fix severity:

**Must-fix**
- `walk.ts`: symlink-escape + crash risk — `walkProject` followed symlinks with no containment check or cycle detection, which could both bypass the `projects` allowlist and crash the whole gateway process via unbounded recursion. Fixed with `realpathSync` containment checks and a visited-directories cycle guard (seeded with the root itself, since a naive per-child guard still let a *self*-referential top-level symlink slip through once). Added `src/__tests__/walk.test.ts` covering escape, deep-cycle, and self-cycle cases with real filesystem symlinks.
- `dashboard-route.ts`: race condition — `getOrStartViewer` only cached a viewer once it had fully started (up to 15s), so two concurrent dashboard requests for the same project both spawned a viewer, leaking one as an orphaned process. Fixed by caching the in-flight promise synchronously. Verified live (not just by inspection): fired two concurrent requests against the built module and confirmed both resolved to the identical viewer instance.
- No lifecycle management for spawned viewer processes — added a `gateway_stop`-hooked shutdown plus a 30-minute idle-eviction sweep.
- `llm.ts`: no timeout on the Anthropic HTTPS call, so one stalled connection could wedge an analysis job in `"running"` state permanently. Added a 60s request timeout.

**Should-fix:** non-atomic graph writes could throw on concurrent reads (now falls back to cached graph), the API-key fallback no longer silently defaults to reading the `"main"` agent's auth profile in multi-agent deployments, defensive field access on graphs loaded with `validate:false`, and a missing `?project=` query param no longer silently resolves to index 0.

**Nice-to-have:** removed a redundant duplicate `builder.build()` call.

Full pipeline re-verified end-to-end after these changes (real Anthropic key, real project) with no regression — matching node/edge counts, 0 warnings. 17/17 unit tests pass (13 existing + 4 new).

zeroaltitude:
Added the interactive Ask chat panel discussed as a natural extension of this plugin — the dashboard can now be talked to directly instead of needing a CLI/chat session for Q&A.

**What's new:**
- A floating chat widget on the dashboard, backed by a new `POST /ask.json` endpoint.
- `src/ask.ts`: grounds answers in the persisted knowledge graph — `SearchEngine` finds relevant nodes for the question, reads a bounded number of their source files, and a single LLM call (the same caller used for analysis) answers using only that context. Same idea as upstream's `/understand-chat` skill, just reachable from the browser.
- `src/interactive-server.ts`: a **new, separate** server (not a fork of `understand-anything-viewer`) that reuses the viewer's static dashboard build and JSON API routes read-only, and adds the widget + `/ask.json`. `understand-anything-viewer` itself is untouched — its whole reason to exist is staying LLM-free for team-sharing, so that path is deliberately never modified.
- `dashboard-route.ts` now serves the interactive server when an API key is configured, and falls back to the plain zero-LLM viewer exactly as before when it isn't — both paths verified live.

**Verification:**
- Live end-to-end test with a real Anthropic key against a real analyzed project: the widget serves, `/ask.json` answers a real question accurately (correctly identified and cited the actual multi-file ingestion code paths), existing graph/file-content API unaffected.
- No-API-key fallback path re-verified: still returns the plain viewer with no widget injected.
- 3 new unit tests for `ask.ts` (no-graph case, node-matching/grounding, snippet inclusion) — 20/20 total tests pass.
- Same token/security model throughout: `/ask.json` requires the per-instance access token (sent as `X-Ask-Token` since it's a POST body) and only reads files already listed in the graph, same constraint as `/file-content.json`.

zeroaltitude:
Two more picker-page additions:

**1. Click to analyze.** Unanalyzed projects get an "Understand this project" button right on the picker page instead of a "call the tool" note — shares one job-tracking path with the `understand_analyze_project` tool (extracted a `startAnalysis()` function rather than a second implementation). Auto-refreshes while analysis is running, shows "Retry analysis" if a run failed.

**2. Add a project from the dashboard**, gated behind a new `allowAddProject` config flag (default **false**): paste a GitHub URL or an existing local path and it's added + analysis starts immediately.

- `src/project-store.ts`: combines config-declared projects (fixed) with dynamically-added ones (persisted to disk so they survive a restart) into one index-stable list.
- GitHub handling only recognizes an actual `https://github.com/...` or `git@github.com:...` URL — deliberately not a bare `owner/repo` shorthand (ambiguous with a real local path) or any non-GitHub/`file://` URL (which could otherwise smuggle an arbitrary local path into a "clone"). Shallow clone, 3-minute timeout, real array-args `spawn()` (no shell).
- `allowAddProject` defaults off and is documented as a genuine security tradeoff: the `projects` allowlist is otherwise the only way to expand what this plugin can read/analyze/serve.

**Verified end-to-end with a real live GitHub clone**, not just unit tests: activated the full plugin, clicked "Understand this project" on a fresh project (job started, confirmed via `understand_status`), then added `octocat/Hello-World` by its real URL — genuinely cloned, appeared in the picker and `understand_list_projects` at the correct index, auto-started analysis. Also confirmed the `allowAddProject: false` default correctly hides the form and rejects the endpoint.

6 new tests for `ProjectStore` — 26/26 total tests pass.

zeroaltitude:
Three more additions, plus a real bug fix uncovered while testing:

**1. Re-analyze button** on already-analyzed projects (analysis was always a full fresh re-run — no UI trigger existed for that case).

**2. Auto-generated tours at analysis time.** A free, deterministic module walkthrough (wraps core's `generateHeuristicTour`) syncs into the standard `graph.tour` field, so upstream's Learn persona/LearnPanel keeps working with zero changes. A second, LLM-narrated **code-review tour** ranks nodes by complexity + how central they are in the dependency graph, then explains *why* a reviewer should care about each — not just what the code does. Both persist to a plugin-owned `.ua/tours.json` sidecar, since the schema only supports one named tour. (Lifecycle/request-flow tours are explicitly deferred — that needs call-graph/endpoint traversal work.)

**3. Custom tour generation** + a new Tours widget (same vanilla-JS pattern as the Ask widget): select node(s) in the live graph canvas — read via a `MutationObserver` on `.react-flow__node.selected`, zero patches to the dashboard's React source — type a prompt, get back a tour scoped to just that selection and request.

**Bug fix found along the way:** `llm.ts`'s response parsing assumed `content[0]` was always the text block, but extended thinking (on by default for some models) puts a `"thinking"` block first, with the real response in a later block. Every LLM call in the plugin shared this latent bug — it just hadn't been triggered before. Fixed to find the first `type: "text"` block instead of blindly indexing `[0]`, with 4 regression tests mocking `node:https`.

**Verified end-to-end with real live Anthropic calls:** analyzed a real project (both tours generated correctly, zero warnings post-fix), started the interactive server, confirmed both widgets inject, hit the tours API, then generated a custom tour from real node ids + a real prompt — got back an accurate 5-step walkthrough correctly tracing the actual code path. All validation/security paths confirmed.

14 new tests — 44/44 total pass.

zeroaltitude:
Fourth addition: **PR/diff walkthroughs**.

Answers "can this framework understand a *change*, not just a static codebase?" — yes: `understand_analyze_pr` (agent tool) and a new "Generate PR walkthrough" control in the Tours widget resolve a PR's (or branch diff's) changed files via `gh pr diff` or `git diff <base>...HEAD` (falling back to uncommitted working-tree changes), compute the blast radius against the already-analyzed graph with a new pure, deterministic `computeDiffOverlay` in `@understand-anything/core` (changed files → file/function/class nodes sharing the same `filePath`, then a 1-hop edge walk both directions — no LLM involved in this step), and only then ask the model to narrate a walkthrough grounded strictly in the changed + affected nodes it's given. The result — overlay JSON + generated tour — persists the same way as a custom tour (`kind: "prWalkthrough"` in `.ua/tours.json`), and the overlay itself is saved to `.ua/diff-overlay.json`, formalizing the on-disk contract the `understand-diff` skill already documents informally.

This one touched `@understand-anything/core` (the new `DiffOverlay` type + `saveDiffOverlay`/`loadDiffOverlay` persistence + `computeDiffOverlay`), so it's split across two branches: `feat/diff-overlay-core` (core, opened as a separate PR — logically distinct contribution) merged into this branch to build against.

**Bug found + fixed via live E2E testing, not just unit tests:** the LLM call's `max_tokens: 2000` was fine for small diffs but truncated mid-JSON on a real diff touching 18 functions in one file (`stop_reason: "max_tokens"`), producing invalid JSON the parser correctly rejected rather than silently returning garbage. Raised the budget to 8000 and capped how many changed/affected nodes get listed in the prompt (40/20) so required output stays bounded regardless of diff size — large diffs now tell the model how many nodes were omitted and ask it to summarize rather than enumerate. Re-verified against the same real diff afterward: clean 7-step walkthrough, no truncation.

5 new tests (real git-repo integration tests for changed-file resolution + stub-LLM grounding tests) — 49/49 total pass.

zeroaltitude:
Fifth addition: **Ask + Tours now ground on the user's live graph selection, through one shared mechanism.**

Prompted by watching it fail in practice: selected a file via the dashboard's Files breadcrumb, asked the Ask panel a question, and the answer never referenced the selected file at all — because `askAboutProject` had zero notion of what the user was looking at, unlike custom-tour generation (which already scoped to an explicit node selection).

- Extracted the selection-reading logic (previously private to `tours-widget.js`, a `MutationObserver` on `.react-flow__node.selected`) into a new shared `src/selection.js`, loaded once before both widgets — `window.uaSelection.get()`/`.subscribe()`. One observer instead of two independently watching the same DOM.
- `askAboutProject` gained an optional `selectedNodeIds` param: selected nodes are always included as primary grounding context (deduped against `SearchEngine`'s free-text matches), ahead of and independent from whatever the question's wording happens to match. The Ask panel now shows a "Grounded in: ..." bar so it's visible whether a selection was actually picked up.

**Verified live, not just unit-tested:** asked the identical vague question ("what does this file do?") against a real analyzed project twice — once with nothing selected (answered about whatever ranked highest generically) and once with `src/db.ts` selected (correctly answered about `db.ts` specifically) — same question, different grounding, correct behavior both times.

3 new tests — 52/52 total pass.

zeroaltitude:
Closing in favor of #575 — same content, reopened from a correctly-named branch (`feat/openclaw-gateway-plugin` instead of `pr/openclaw-gateway-plugin`) as part of a branch-naming cleanup in our fork. All prior discussion/history here carries forward via #575's description.

PR #571, opened 2026-07-12
core: add DiffOverlay type + persistence + blast-radius computation

zeroaltitude:
Formalizes an on-disk contract that already existed informally: the `understand-diff` skill has always read/written `<ua-dir>/diff-overlay.json` in this exact shape (SKILL.md step 8), and the dashboard's store/GraphView already consume `changedNodeIds`/`affectedNodeIds` for the diff-mode overlay UI — but core had no first-class type or persistence pair for it, unlike the sibling `domain-graph.json` (`saveDomainGraph`/`loadDomainGraph`).

- `DiffOverlay` type in `types.ts`.
- `saveDiffOverlay`/`loadDiffOverlay` in `persistence/index.ts`, mirroring the domain-graph save/load pattern exactly (same directory resolution, same file-path sanitization so absolute paths never leak to disk — extracted the per-path sanitization logic out of `sanitiseFilePaths` into a reusable `sanitiseFilePath` so both the node-array and plain-string-array shapes can share it, no behavior change to the existing function).
- `computeDiffOverlay(graph, changedFiles, baseBranch?)` in a new `diff-overlay.ts` — pure and deterministic: maps changed file paths to file/function/class nodes (graph-builder stamps the same `filePath` on all three), then walks 1-hop edges in both directions for the blast radius, excluding the changed nodes themselves.

11 new tests (5 persistence round-trip/sanitization, 6 traversal logic). Full suite: 942/942 pass.

This is a companion PR to #569 (the OpenClaw gateway plugin), which consumes this to add PR/diff-walkthrough tours — but the type/persistence/traversal logic here is general-purpose and useful independent of that integration, so splitting it out as its own contribution.

### DIFF
--- a/understand-anything-plugin/packages/core/src/__tests__/diff-overlay-persistence.test.ts
+++ b/understand-anything-plugin/packages/core/src/__tests__/diff-overlay-persistence.test.ts
@@ -1,0 +1,63 @@
+import { describe, it, expect, beforeEach, afterEach } from "vitest";
+import { mkdirSync, rmSync, existsSync, writeFileSync } from "node:fs";
+import { join } from "node:path";
+import { tmpdir } from "node:os";
+import { saveDiffOverlay, loadDiffOverlay } from "../persistence/index.js";
+import type { DiffOverlay } from "../types.js";
+
+const testRoot = join(tmpdir(), "ua-diff-overlay-persist-test");
+
+const overlay: DiffOverlay = {
+  version: "1.0.0",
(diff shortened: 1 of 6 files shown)


zeroaltitude:
Closing in favor of #574 — same content, reopened from a correctly-named branch (`feat/diff-overlay-core` instead of `feat/diff-overlay-core-upstream`) as part of a branch-naming cleanup in our fork.

PR #572, opened 2026-07-12
feat: ground Ask + Tours on live graph selection via a shared discovery script

zeroaltitude:
Branches off `integration` (this repo's aggregation branch for the OpenClaw gateway plugin work in #569/#571), so this diff is just the new selection-grounding feature on top of what's already there.

Ask (`src/ask.ts`) previously had zero notion of what the user is currently looking at — pure free-text `SearchEngine` relevance. Custom-tour generation already scoped to an explicit node selection (read via a private `MutationObserver` on `.react-flow__node.selected` in `tours-widget.js`), so a question asked while a specific file was selected in the canvas could ignore that file entirely if the question's wording didn't happen to match it.

- Extracted the selection-reading logic into `src/selection.js`, a single shared discovery script (`window.uaSelection.get()`/`.subscribe()`) loaded before both widgets, so there's exactly one `MutationObserver` instead of two independently watching the same DOM. Both `ask-widget.js` and `tours-widget.js` now read from it instead of keeping private copies.
- `askAboutProject` gained an optional `selectedNodeIds` param: selected nodes are always included as primary grounding context (deduped against `SearchEngine`'s matches), ahead of and independent from whatever the question text happens to match. The Ask panel also shows a "Grounded in: ..." bar so it's visible whether a selection was picked up.

Verified live against a real analyzed project: the same vague question ("what does this file do?") answered about whatever file ranked highest generically without a selection, and correctly answered about the selected file (`src/db.ts`) once one was set — despite an identical question in both cases.

3 new tests (selection grounding, selected+matched dedup, unknown node id ignored) — 52/52 total pass.

### DIFF
--- a/.openclaw-plugin/README.md
+++ b/.openclaw-plugin/README.md
@@ -23,9 +23,21 @@ connected agent can call mid-conversation, plus a dashboard route.
 **HTTP routes** (mounted on the gateway):
 
 - `GET /understand-anything` — project picker
-- `GET /understand-anything/open?project=<idx>` — starts (or reuses) a
-  `understand-anything-viewer` instance for that project and redirects to its
-  token-protected dashboard URL. The viewer binds to 127.0.0.1 only.
+- `GET /understand-anything/open?project=<idx>` — starts (or reuses) a viewer
+  instance for that project and redirects to its token-protected dashboard
+  URL. Binds to 127.0.0.1 only.
+
+  With an Anthropic API key configured, this serves **interactive-server.ts**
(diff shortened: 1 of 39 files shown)


zeroaltitude:
Closing — this depends on the whole plugin scaffold from #569 (ask.ts, tours-widget.js, etc.), so it can't stand alone against `main` the way #571 could (that one was genuinely core-only and cherry-picked cleanly onto a fresh base). Re-proposing this as an addition to #569 instead, same as every other feature added there over time.

PR #585, opened 2026-07-16
feat: add deterministic no-LLM code graph

tkhjp:
## Summary

- add an `/understand-no-llm` skill for deterministic code-graph generation
- reuse the existing scanner, import resolver, Tree-sitter extractors, and KnowledgeGraph validation
- emit file/function/class nodes and contains/imports/calls edges
- add Japanese documentation covering architecture, file responsibilities, usage, security boundaries, and limitations

## Why

Some environments cannot send source code to an LLM but still need a navigable structural code graph. This adds a narrower execution path that performs local static analysis only while keeping the normal KnowledgeGraph JSON shape for dashboard compatibility.

## Behavior and scope

- does not invoke LLM agents or external AI APIs
- does not execute target repository build, test, application, or shell scripts
- resolves calls conservatively only when caller and callee are unambiguous
- writes `code-graph.json` and a coverage/report file
- leaves `layers` and `tour` empty and does not claim semantic parity with `/understand`
- can optionally copy the result to `knowledge-graph.json` for dashboard use

## Validation

- `node --check understand-anything-plugin/skills/understand/no-llm-code-graph.mjs`
- `pnpm --filter @understand-anything/core test`: 35 files, 753 tests passed
- smoke-tested against `soongenwong/claudecode` with `--no-scripts`: 119 code files, 1,800 nodes, 2,578 edges, 0 skipped structure files


Lum1104:
Hi, the idea is great. How about we add this to the `/understand` skill, and let agents pause after the non-llm graph generation? Or, at the very beginning, during user's interaction with .understand ignore?

PR #576, opened 2026-07-13
fix(dashboard): move force layout off the main thread

KumamuKuma:
## Summary

Move the dashboard knowledge graph force layout from the main thread into a dedicated Worker. Superseded layout jobs are terminated, stale request IDs are ignored, ReactFlow mounts after positions are ready, and Worker failures use an O(N) grid fallback instead of running d3-force on the main thread.

The force-layout algorithm also normalizes sparse community IDs (for example, `0` and `2`) to contiguous indices before computing cluster angles, so empty layers cannot collapse distinct communities onto the same center.

Closes #491

## Relationship to #490

PR #490 intentionally left the synchronous d3-force Worker migration to Issue #491 as a separate follow-up. This PR implements that follow-up on top of a main branch that already contains #490; it does not duplicate #490's other hot-path optimizations.

## Before / after

| 3,000 nodes / 6,000 edges | Before | After |
| --- | ---: | ---: |
| Main-thread max 10 ms timer gap | 4,550.4 ms | 11.3 ms |
| Total layout wall time | 4,549.9 ms | 4,214.3 ms |
| Real-dashboard max Long Task | 9,701 ms | 1,795 ms |
| Real-dashboard layout-ready time | 13,020.8 ms | 7,765.8 ms |

The key fix is responsiveness: dispatch returned in 1.7 ms while the Worker produced all 3,000 node positions.

## Testing

- [x] pnpm lint
- [x] core/skill/viewer/dashboard builds
- [x] core tests: 931 passed
- [x] dashboard tests: 67 passed
- [x] pnpm test
- [x] Python helper tests: 95 passed, 1 expected Windows skip
- [x] Manual Chrome smoke test: deterministic 3,000-node / 6,000-edge graph rendered all nodes

## Versioning

- [x] Version bumped across all six manifests: 2.9.2 -> 2.9.3

## Detailed report

[Issue #491 Worker layout performance and validation report (PDF)](https://github.com/user-attachments/files/29965733/understand-anything-issue-491-worker-layout-report.pdf)

Issue #483, opened 2026-06-19
Investigation: evaluate Understand-Anything for onboarding onto large codebases


xmanjack:
## Context

We've added Understand-Anything as a git submodule in our research repo and want to evaluate it as a tool for quickly understanding large/unfamiliar codebases via its knowledge-graph pipeline.

## Goals of this investigation

- [ ] **Pipeline & graph quality** — Understand how the multi-agent pipeline analyzes a project and how accurately the knowledge graph captures files, functions, classes, and dependencies.
- [ ] **Scale & cost** — Evaluate behavior, runtime, and token/cost on a large repo (100k+ LOC), plus any incremental/re-analysis support.
- [ ] **Integration & dashboard** — Assess the Claude Code plugin install flow, cross-tool support (Codex, Cursor, Copilot, Gemini CLI), and the interactive dashboard's usefulness for exploration and Q&A.

## Questions to answer

- How well does it handle our primary languages/frameworks?
- What data leaves the machine, and what are the privacy/security implications?
- How does graph accuracy degrade on very large or polyglot repos?

## Deliverable

A short write-up summarizing findings and a go/no-go recommendation.

Lum1104:
Sounds good, I am insterested in large scale evaluation too. Can you scale more, like to 1000k loc, or a monorepo? Feel free to contact me at lin.yuxiang.contact@gmail.com

wusy11122132:
蹲，目前生成的图谱作为新人查看理解成本还是比较高，可以给出些教程实例吗

PR #587, opened 2026-07-17
perf(bench): add reproducible large-monorepo benchmark and scale report

KumamuKuma:
## Summary

This PR adds a reproducible benchmark for the non-LLM stages of `/understand`:
file scanning, import extraction, batch planning, and Tree-sitter structure
extraction.

It gives contributors a shared way to measure large repositories and find
bottlenecks using repeatable evidence. It does not call an LLM or write
analysis data into the repository being tested.

Each run produces matching JSON and Markdown reports with repository scale,
timing, CPU and memory use, warnings, failures, and input/output digests.
Private paths, hostnames, and Git remotes are not included.

## Usage

```sh
corepack pnpm benchmark:large-repo /absolute/path/to/repository \
  --label public-repository-name \
  --output /absolute/path/to/results/report.json
```

The output must be outside the tested repository. Temporary files are cleaned
up by default.

## What this adds

- Deterministic file selection, ordering, batching, and content digests.
- Isolated benchmark stages with bounded concurrency and failure reporting.
- Paired JSON/Markdown reports validated by schema version `1.0.0`.
- Cross-platform instructions and a pinned TensorFlow reproduction recipe.

## Related work

- Refs #76 for `/understand` performance work.
- Refs #483 for large-codebase evaluation.
- Follow-up to the scale discussion around #576.

## Limitations

- This is not an end-to-end `/understand` benchmark. It does not run an LLM,
  count tokens, estimate cost, build the final knowledge graph, or open the
  dashboard.
- `estimatedAgentInputBytes` measures bytes, not tokens or cost.
- TensorFlow is currently a pinned recipe, not a completed measurement.
- This PR does not claim verified support for one-million-line repositories.
- Timing and memory should only be compared across same-machine runs with the
  same concurrency and matching digests.

## Current plan

1. Land the reproducible benchmark and report format.
2. Collect at least three same-digest community runs, including the pinned
   TensorFlow revision and other monorepos.
3. Compare repeated runs on the same machine.
4. Use the results to choose and verify the next optimization.
5. Benchmark LLM, token, and cost behavior separately in future work.

## Testing

- Install, lint, and core/skill/viewer builds passed.
- Core tests: 934 passed.
- Root Vitest suite: 380 passed, 7 platform-specific skips.
- Python helpers: 95 passed, 1 platform-specific skip.
- Benchmark/schema tests: 70 passed, 1 platform-specific skip.
- A real CLI smoke test produced schema-valid paired reports, leaked no private
  paths, and left the tested repository clean.

## Versioning

N/A — this PR adds internal benchmark tooling, tests, and documentation.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@ dist
 .env
 .env.*
 coverage/
+benchmark-results/
 *.log
 __pycache__/
 .claude/
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -110,6 +110,13 @@ Then open a Pull Request on GitHub with:
 
 ## 🧪 Testing Guidelines
 
+### Large Repository Benchmarks
+
+Use the [large monorepo benchmark](docs/benchmarks/large-monorepo.md) to collect
+reproducible scale, timing, and memory reports from projects that are too large
+for CI fixtures. The guide includes a pinned TensorFlow example and explains
+which deterministic stages are measured.
+
 ### Writing Tests
 
 - Use Vitest for testing
--- a/docs/benchmarks/large-monorepo.md
+++ b/docs/benchmarks/large-monorepo.md
@@ -1,0 +1,146 @@
+# Large Monorepo Benchmark
+
+Use this benchmark to collect reproducible scale and performance evidence from
+large repositories without running an LLM. It is intended for local and
+community runs on projects that are too large for normal CI fixtures.
+
+## Scope
+
+The runner executes the real deterministic helpers used by `/understand`:
+
+1. file scanning, classification, and line counting;
+2. static import-map extraction;
+3. semantic batch planning; and
+4. Tree-sitter structural extraction for every batch.
+
+It reports wall-clock time, CPU time where available, peak resident memory,
+output size, repository scale, batching statistics, structural coverage, and
+deterministic SHA-256 digests.
+
+For the regular scan, imports, and batching helpers, `peakRssBytes` is the peak
+RSS of that helper process. The concurrent structure stage instead reports
+`maxWorkerPeakRssBytes`, the maximum peak RSS of any individual worker, not a
+sum across workers. Its `userCpuTimeMicros` and `systemCpuTimeMicros` values are
+summed across all structure workers.
+
+It does **not** run an LLM, call an API, count tokens, estimate cost, generate a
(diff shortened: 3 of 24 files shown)


thejesh23 (commented):
@KumamuKuma  Can you please share the sample report to manual review, it would be easier to visualize the benchmarking result expectation through this PR. If you already have one for tensor-flow you can share  that as well

KumamuKuma:
@thejesh23 Thanks for the suggestion. I added full-repository sample reports for manual review:

- [Markdown report](https://github.com/KumamuKuma/Understand-Anything/blob/48b4fe474b0b5acf993616c38e350da535185001/docs/benchmarks/samples/understand-anything-pr587-full-sample.md)
- [JSON report](https://github.com/KumamuKuma/Understand-Anything/blob/48b4fe474b0b5acf993616c38e350da535185001/docs/benchmarks/samples/understand-anything-pr587-full-sample.json)

The sample covers this repository itself: 459 files and 127,155 lines. It shows repository scale, stage timing, CPU and memory usage, warnings, and integrity information for the deterministic non-LLM stages. The JSON validates against schema 1.0.0, and its pair ID matches the Markdown report.

I have not generated a TensorFlow report locally because its size requires more time and resources than I can reliably provide in my current setup. The pinned TensorFlow reproduction recipe remains available in this PR for community testing.

While preparing the sample, I noticed a small report-classification issue, so I added a follow-up commit to document the expected behavior, correct the issue, strengthen result validation, and include the final sample reports.

KumamuKuma:
@thejesh23 Thanks again for reviewing this.
My next plan is to collect more same-digest benchmark runs from large repositories, especially the pinned TensorFlow revision. I would then compare repeated runs made on the same machine with the same concurrency, identify which stage is consistently the main bottleneck, and use that evidence to choose and verify the next optimization. LLM, token, and cost measurements would remain a separate follow-up.
If you have time, would you be willing to help run the pinned TensorFlow benchmark and share the paired Markdown and JSON reports? The tool records the repository commit, environment, digests, timing, memory usage, and integrity results, so the reports can be compared reliably. A run from TensorFlow—or another large monorepo—would be very helpful.

thejesh23:
@KumamuKuma  Sure I can and I will . Will do it after I'm back from vacation next week.

PR #609, opened 2026-07-28
fix(hooks): emit PostToolUse additional context

KumamuKuma:
## Summary

- extract the inline PostToolUse auto-update logic into a portable Node handler
- emit `hookSpecificOutput.additionalContext` so the instruction reaches model context
- preserve the existing Git-operation, config, graph, and `.ua` fallback guards
- add regression coverage for successful triggers and silent failure paths

## Context

Follow-up to #598. That PR fixed the hook input by reading the PostToolUse payload from stdin, but the handler still wrote a plain-text message to stdout. PostToolUse ignores plain stdout, so the commit-triggered auto-update instruction was never added to model context.

## Validation

- `pnpm lint`
- `pnpm --filter understand-anything-viewer build`
- `pnpm test` (new hook tests: 14/14)
- Python helper suite: 95 tests, 1 skipped
- `git diff --check origin/main...HEAD`

### DIFF
--- a/tests/hooks/post-tool-use-auto-update.test.mjs
+++ b/tests/hooks/post-tool-use-auto-update.test.mjs
@@ -41,13 +41,23 @@ function runHook({
 }
 
 describe('PostToolUse auto-update hook', () => {
-  it('emits the auto-update instruction after a git commit', () => {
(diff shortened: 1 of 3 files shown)


Lum1104:
@codex review this

PR #613, opened 2026-07-30
fix: ignore braces inside strings and comments when matching blocks

signalblur:
## Summary

The terraform, protobuf, graphql, and shell parsers find a block's extent by counting `{` and `}` as raw characters, so a brace inside a string literal or a comment decrements the depth early and truncates the block. This adds a shared brace matcher that tracks quote and comment context, and fixes a separate GraphQL bug where `extractDefinitions` did no depth counting at all.

## What was wrong

The truncation costs more than line numbers. `extractMessageFields` returns an empty array for a protobuf message whose first field carries an option with a brace in its value, and `extractFields` drops GraphQL fields that follow a default value containing one.

`extractDefinitions` in `graphql-parser.ts` used `indexOf("}")` with no depth tracking, so the first brace in the file ended every definition regardless of nesting.

## The fix

`brace-matcher.ts` holds the state machine, parameterized by a per-language `BraceSyntax` naming the quote and line-comment forms to honor. `findClosingBrace` returns an index for the parsers that slice by offset; `countBracesPerLine` returns per-line deltas for the shell parser, which walks lines. Unbalanced input still returns `content.length` and warns under the calling parser's name, matching the four private implementations it replaces.

The `extractDefinitions` scan now runs only when a definition's header line opens a brace, so a bodyless `scalar` or `union` no longer matches a later definition's closing brace.

## Linked issue(s)

None.

## How I tested this

11 regression tests in `parsers.test.ts`: string and comment cases per parser, an escaped-quote case, a backslash-continued line case, and the bodyless-scalar case. Reverting the fix fails exactly those 11 and leaves the 61 pre-existing parser tests green, so each one pins the defect rather than the implementation.

Expected values came from each language's own toolchain, not from reading the parser: `bash -n` plus execution, `terraform fmt -check -diff`, `buf build`, and `graphql@16`'s `parse()`, which reports lines 1-4 for the case the parser reported as 1-3.

`tsc --noEmit` clean. Core suite 987 passed. Root suite 496 passed, 12 skipped.

One unrelated note: `tests/benchmark/test_large_repo_benchmark.test.mjs:253` asserts the benchmark output contains no `/w`, so it fails when the repo sits at a path that matches that substring. Passes at a normal checkout path.

- [x] `pnpm lint`
- [x] `pnpm --filter @understand-anything/core test`
- [x] `pnpm test`
- [x] Manual smoke test — ran all four parsers over standalone `.tf`, `.proto`, `.graphql`, and `.sh` fixtures outside the suite, comparing output against the reference toolchains above

## Versioning

- [ ] Version bumped in all five manifests, OR
- [x] N/A — internal fix, no API change; leaving the bump to the maintainer on merge to avoid conflicting with other open PRs

### DIFF
--- a/understand-anything-plugin/packages/core/src/__tests__/parsers.test.ts
+++ b/understand-anything-plugin/packages/core/src/__tests__/parsers.test.ts
@@ -627,3 +627,129 @@ describe("registerAllParsers", () => {
     expect(registry.getSupportedLanguages()).toContain("shell");
   });
 });
+
+describe("brace matching ignores strings and comments", () => {
+  it("terraform: a closing brace in a string does not end the block", () => {
+    const content = `resource "aws_ssm_parameter" "example" {
+  name  = "/app/config"
+  value = "suffix-}-marker"
+  type  = "String"
+}`;
+    const result = new TerraformParser().analyzeFile("main.tf", content);
+    expect(result.resources![0]).toMatchObject({ name: "aws_ssm_parameter.example" });
+    expect(result.resources![0].lineRange).toEqual([1, 5]);
+  });
+
+  it("terraform: a closing brace in a comment does not end the block", () => {
+    const content = `variable "example" {
+  # a brace in a comment: }
+  type    = string
+  default = "x"
+}`;
+    const result = new TerraformParser().analyzeFile("main.tf", content);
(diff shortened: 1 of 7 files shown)


Lum1104:
@codex review this

PR #651, opened 2026-08-17
feat(understand-diff): optional graph query backend for retrieval

galshubeli:
## What this adds

An **opt-in, read-only** query adapter over the existing `knowledge-graph.json`, plus an optional fast path in `understand-diff` that uses it. Follows up on discussion #649.

Nothing changes by default. The JSON stays the source of truth, nothing is written back, and when no backend is configured the skill continues with the existing grep steps exactly as before. The diff is additive: no existing line is modified or removed.

## Why

`understand-diff` resolves a changed file's blast radius by grepping `knowledge-graph.json` for node ids, then grepping again for each id in the edges array. That's a multi-hop traversal done as repeated text search, so its cost grows with the size of the graph rather than the size of the answer.

Measured on this repo's own graph (793 nodes / 1390 edges, produced by `/understand`), for one changed file (`packages/core/src/staleness.ts`):

| | context consumed | answer |
|---|---|---|
| current grep path (steps 4–5) | ~8,517 tokens | 11 nodes |
| adapter, one `batch` call | ~638 tokens | 11 nodes |

Same 11 nodes, 13x less context. The gap widens with how connected the changed file is: for `packages/core/src/types.ts`, which 41 other nodes import, the grep walk pulls ~107k tokens versus ~2.3k.

Grep is charged only the characters it emits, so the real difference is larger — the model's reasoning tokens between hops aren't counted. The JSON being pretty-printed is the best case for grep.

## How it works

`skills/understand-diff/graph-query.py` mirrors the JSON into a graph and answers queries against it:

```bash
python "<SKILL_DIR>/graph-query.py" batch --q '[
  {"op": "nodes-for-file", "path": "src/changed.ts"},
  {"op": "blast-radius",   "name": "changed.ts", "hops": 3}
]'
```

`nodes-for-file` returns the file node plus every function and class defined in it (step 4). `blast-radius` returns what transitively depends on it (step 5).

Backends are detected automatically:

* **embedded** — [FalkorDBLite](https://github.com/FalkorDB/falkordblite), in-process, no server and no configuration. Requires Python >= 3.12.
* **server** — any FalkorDB instance via `UA_FALKORDB_URL`.

The graph is rebuilt only when the JSON's content hash changes, so repeated queries pay the load cost once. Loading is a script, not an agent, so it costs no LLM tokens.

## Scope

* Doesn't touch the analysis phase, where the real token cost lives.
* Doesn't touch the dashboard, which needs the whole graph in memory to render.
* Only `understand-diff` is wired up here. `chat`, `explain` and `onboard` use the same grep pattern and could follow if you like the shape of this.

## Notes for review

* Process startup dominates a single query (queries are 1-3 ms; the embedded server's boot and shutdown are not). The skill instructions tell it to send all changed files in one `batch` call for that reason.
* The adapter currently lives beside the skill that uses it. If more skills adopt it, it should probably move somewhere shared.
* Happy to reshape this — if you'd rather it sit behind `persistence/index.ts` as a `saveGraph`/`loadGraph` style interface than as a skill-level script, say so and I'll rework it.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -61,4 +61,4 @@ jobs:
         run: pnpm test
 
       - name: Test Python skill helpers
-        run: python -m unittest tests.skill.understand.test_merge_batch_graphs tests.skill.understand.test_merge_subdomain_graphs tests.skill.knowledge.test_parse_knowledge_base -v
+        run: python -m unittest tests.skill.understand.test_merge_batch_graphs tests.skill.understand.test_merge_subdomain_graphs tests.skill.knowledge.test_parse_knowledge_base tests.skill.diff.test_graph_query_helpers tests.skill.diff.test_graph_query -v
--- a/tests/skill/diff/test_graph_query.py
+++ b/tests/skill/diff/test_graph_query.py
@@ -205,6 +205,19 @@ class TraversalTests(GraphQueryTestCase):
                                         "hops": 1}]))
         self.assertEqual(out[0], ["file:src/a.ts", "file:src/b.ts"])
 
+    def test_search_honours_an_explicit_limit(self) -> None:
+        """A limit passed in a batch spec must reach the query.
+
+        It used to be dropped, so a caller asking for 2 results silently got 25 --
+        the opposite of what a context-conscious skill wants.
+        """
+        out = self.cli_json("batch", "--q", json.dumps([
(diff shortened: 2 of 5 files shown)


Lum1104:
Haven't got a change to look at the implementation, but do you want to have the committer to you account?

image

PR #691, opened 2026-09-14
fix: correct Hermes skills directory path on Windows

jayv00691-sudo:
## Summary
Hermes stores its data in `%LOCALAPPDATA%\hermes\skills`, not `~/.hermes/skills`. Both `install.sh` and `install.ps1` pointed to the wrong location, causing Understand-Anything to be installed into a directory that Hermes never reads.

## Root Cause
The install scripts assumed Hermes follows the dotfile convention (`~/.hermes/`), but on Windows Hermes uses the AppData directory. The two scripts (bash + PowerShell) both had the same bug — suggesting the PowerShell version was adapted from the bash version without accounting for the different data location.

## Changes
- **install.sh**: `$HOME/.hermes/skills` → `$HOME/AppData/Local/hermes/skills`
- **install.ps1**: `Join-Path $HOME .hermesskills` → `Join-Path $env:LOCALAPPDATA hermesskills`

## Test Plan
- [ ] Install on Windows using `install.ps1 hermes` and verify the symlink points to `%LOCALAPPDATA%\hermes\skills\understand-anything`
- [ ] Install on macOS/Linux using `install.sh hermes` and verify it still creates the symlink at the correct path
- [ ] Verify Hermes loads the understand-anything skill after restart


Lum1104 (changes requested):
See the comment~

